PR #1, opened 2023-04-12
Update star badge.

wu-sheng:


mrproliu (approved):

wu-sheng:
@mrproliu I think you missed a CI task when you submit to the `protected_branches`.

heyanlong (approved):

PR #17, opened 2023-04-26
Change to using workspace

mrproliu:
In this PR, I have down the following things:
1. Change to using the Go workspace
2. Add some test cases
3. Export the build directory when debugging the test of the plugin
4. Some code suggesting from @jcchavezs (please help take a look on this PR)


wu-sheng (approved):

plugins/core/propagation_test.go
```diff
@@ -0,0 +1,240 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tc := &SpanContext{}
+			if err := tc.DecodeSW8(tt.args.header); (err != nil) != tt.wantErr {
+				t.Errorf("DecodeSW8() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if tt.fields != nil {
+				if tc.Sample != tt.fields.Sample {
```
jcchavezs, line 91: You can save a lot of time and do better assertions by using something like `assert`, see https://pkg.go.dev/github.com/stretchr/testify/assert

Also in this case you just fail but don't tell why so it is hard to debug.

plugins/core/propagation_test.go
```diff
@@ -0,0 +1,240 @@
...
+		want string
+	}{
+		{
+			name: "empty",
+			data: make(map[string]string),
+			want: "",
+		},
+		{
+			name: "empty value",
+			data: func() map[string]string {
```
jcchavezs, line 213: Do you really need a function here? Can't you just create the map directly?

test/plugins/runner-helper/context.go
```diff
@@ -59,7 +59,7 @@ func BuildContext() (*Context, error) {
 		ScenarioName: *scenarioName,
 		CaseName:     *caseName,
 		GoAgentPath:  filepath.Clean(*goAgentPath),
-		DebugMode:    *debugMode,
+		DebugMode:    *debugMode == "on",
```
jcchavezs, line 62: How about `ON` or `On`?

tools/go-agent/tools/types.go
```diff
@@ -17,10 +17,19 @@
 
 package tools
 
+var basicDataTypes = make(map[string]bool)
```
jcchavezs, line 20: I suggest using `map[string]struct{}` as it is less space but also to declare the map in the var instead of initializing on init.

PR #21, opened 2023-05-04
Add Performance test case

mrproliu:


docs/en/agent/performance-tests.md
```diff
@@ -1,3 +1,94 @@
...
-Currently, this part has not been done yet.
\ No newline at end of file
+Performance testing is used to verify the impact on application performance when using SkyWalking Go.
+
+## Test Objective
+
+By launching both the **agent** and **non-agent** compiled applications, we subject them to the same **QPS** under stress testing, 
+evaluating the CPU, memory, and network latency of the machine during the testing period.
+
+The application has been saved and submitted to the [test/performance](../../../test/performance) directory, with the following topology:
```
wu-sheng, line 10:
```suggestion
The application has been saved and submitted to the [test/benchmark-codebase](../../../test/benchmark-codebase) directory, with the following topology:
```

I think this is a better name.

docs/en/agent/performance-tests.md
```diff
@@ -1,3 +1,94 @@
...
+
+```
+pressure -> consumer -> provider
+```
+
+The stress testing service (pressure) uses multithreading to send HTTP requests to the consumer service.
```
wu-sheng, line 16:
```suggestion
The payload(traffic) generator uses multithreading to send HTTP requests to the consumer service. 
```

docs/en/agent/performance-tests.md
```diff
@@ -1,3 +1,94 @@
...
+This consumes a certain amount of CPU usage, preventing idle processing during service stress testing and amplifying the impact of the Agent program on the business application.
+
+### Stress Testing Service
+
+We use the [Vegeta](https://github.com/tsenart/vegeta) service for stress testing, which launches traffic at a specified QPS to the application. 
+It is based on the Go language and uses goroutines to provide a more efficient stress testing solution.
+
+## Test Environment
+
+A total of 4 GCP machines are launched, all instances are 4C8G.
```
wu-sheng, line 34:
```suggestion
A total of 4 GCP machines are launched, all instances are running on tbe 4C8G VM.
```

docs/en/agent/performance-tests.md
```diff
@@ -1,3 +1,94 @@
...
+
+### Preparation Phase
+The preparation phase is used to ensure that all machines and test case preparations are completed.
+
+#### Pressure
+Install the [Vegeta service](https://github.com/tsenart/vegeta#install) on the stress testing instance and create the following file(`request.txt`) to simulate traffic usage.
+
+```
+GET http://${CONSUMER_IP}:8080/consumer
+Sw8: 1-MWYyZDRiZjQ3YmY3MTFlYWI3OTRhY2RlNDgwMDExMjI=-MWU3YzIwNGE3YmY3MTFlYWI4NThhY2RlNDgwMDExMjI=-0-c2VydmljZQ==-aW5zdGFuY2U=-cHJvcGFnYXRpb24=-cHJvcGFnYXRpb246NTU2Ng=="
```
wu-sheng, line 53: Is the header changed or not?

docs/en/agent/performance-tests.md
```diff
@@ -1,3 +1,94 @@
...
+```
+
+Please replace the above `CONSUMER_IP` with the real IP address of the consumer instance.
+
+#### Consumer and Provider
+
+Install the skywalking-go service on the machines to be tested, and compile both versions with and without the Agent.
```
wu-sheng, line 60: What do you meab both versions?

wu-sheng (commented):
Generally good, some nits.

docs/en/agent/performance-tests.md
mrproliu: The header does not change.

docs/en/agent/performance-tests.md
mrproliu: just with/without the agent.

wu-sheng: Let's say compile with and without.

mrproliu: Updated.

docs/en/agent/performance-tests.md
wu-sheng: The segment is request thread oriented, right?
So, generally, although all segments are the same trace, it is still good for agent side.

mrproliu: Yes. Although all segments belong to the same trace, the agent must process all headers in all requests.

wu-sheng: OK, got it. Actually, the Java agent doesn't process this part in benchmark.

plugins/http/intercepter.go
```diff
@@ -49,17 +49,11 @@ func (h *Interceptor) AfterInvoke(invocation *operator.Invocation, result ...int
 		return nil
 	}
 	span := invocation.Context.(tracing.Span)
-	res0 := result[0]
-	res1 := result[1]
-	if res0 != nil {
-		resp := res0.(*http.Response)
+	if resp, ok := result[0].(*http.Response); ok && resp != nil {
```
jcchavezs, line 52: How do we know `result` isn't empty?

PR #27, opened 2023-05-09
Support adapt Log frameworks

mrproliu:
Support adapting `logrus` and `zap` log frameworks with the following parts:
1. Integrating Agent logs into the Service
2. Integrating Tracing information into the Service

### DIFF
--- a/docs/en/advanced-features/logging-setup.md
+++ b/docs/en/advanced-features/logging-setup.md
@@ -7,7 +7,7 @@ You can learn about the configuration details through the "log" configuration it
 
 ## Agent with Logging system
 
-The integration of the Agent with logs can be divided into the following:
+The integration of the Agent with logs includes the two parts as following.
 
 1. **Integrating Agent logs into the Service**: Integrating the logs from the Agent into the framework used by the service.
 2. **Integrating Tracing information into the Service**: Integrating the information from Tracing into the service logs.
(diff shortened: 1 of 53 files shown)


docs/en/advanced-features/logging-setup.md
```diff
@@ -1,3 +1,42 @@
...
+
+Agent logs output the current running status of the Agent system, most of which are execution exceptions. 
+For example, communication anomalies between the Agent and the backend service, plugin execution exceptions, etc.
+
+Integrating Agent logs into the service's logging system can effectively help users quickly troubleshoot whether there are issues with the current Agent execution.
+
+### Tracing information into the Service
+
+The Agent would also enhance the existing logging system. 
+When the service outputs log, if the current goroutine contains Tracing data, it would be outputted together with the current logs.
```
wu-sheng, line 25: This reminds me, whether go agent has added the mechanism that would NOT generate any trace information if the backend is not connected.
This is important in the product env. Once the backend(OAP or any others) is down, there is no point to keep tracing.

docs/en/advanced-features/logging-setup.md
```diff
@@ -1,3 +1,42 @@
...
+1. **TraceID**: The current Trace ID. If there is no link, it outputs `Noop`.
+2. **SegmentID**: The Segment ID in the current Trace. If there is no link, it outputs `Noop`.
+3. **SpanID**: The Span ID currently being operated on. If there is no link, it outputs `0`.
+
+The output format is as follows: `[${TraceID},${SegmentID},${SpanID}]`.
+
+The following is an example of a log output when using `Zap.NewProduction`:
+
+```
+{"level":"info","ts":1683635924.717511,"caller":"gin/main.go:45","msg":"test log","SW_CTX":"[6fbe5844ee6611ed8e30acde48001122,6fbe5902ee6611ed8e30acde48001122,0]"}
```
wu-sheng, line 41: Is this the same format as Java agent's SW_CTX? It is better we are always consistent.

docs/en/advanced-features/logging-setup.md
mrproliu: Nope. For now, the gRPC will keep tracing if the connection is closed. I could create other PR for this.

docs/en/advanced-features/logging-setup.md
mrproliu: The format has been updated to the same as the Java agent.

wu-sheng (approved):
LGTM

PR #28, opened 2023-05-09
Fix container name inconsistency in file scenarios.tpl

zheheBao:
When running the run.sh script under skywalking-go/test/plugins, the container name service defined in scenarios.tpl appears_ container_ name，validator_ container_ name.The connection symbol in name is "_", while the container's connection symbol is "-" after docker-compose.yml is started, which causes the scenario. sh runtime to be unable to obtain the corresponding information, resulting in run.sh failure

Before change
Before change

After change
After change

### DIFF
--- a/test/plugins/runner-helper/templates/scenarios.tpl
+++ b/test/plugins/runner-helper/templates/scenarios.tpl
@@ -19,8 +19,8 @@
 set -ex
 
 project_name=$(echo "{{.Context.ScenarioName}}" |sed -e "s/\.//g" |awk '{print tolower($0)}')
-service_container_name="${project_name}_service_1"
-validator_container_name="${project_name}_validator_1"
+service_container_name="${project_name}-service-1"
+validator_container_name="${project_name}-validator-1"
 docker-compose -p "${project_name}" -f "{{.DockerComposeFilePath}}" up -d --build
 
 sleep 3
--- a/test/plugins/runner-helper/templates/scenarios.tpl
+++ b/test/plugins/runner-helper/templates/scenarios.tpl
@@ -18,9 +18,20 @@
 
 set -ex
 
+compose_version=$(docker-compose version --short)
+
+if [[ $compose_version =~ ^(v)?1 ]]; then
+    separator="_"
+elif [[ $compose_version =~ ^(v)?2 ]]; then
+    separator="-"
+else
+    echo "Unsupported Docker Compose version: $compose_version"
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


mrproliu:
Do you use Windows for testing the plugin? I think maybe the `docker-compose` has a different name format for the container. For now, the plugins is works well in Linux and MacOS. 
I think you can judge the system environment and update the name dynamically.

wu-sheng:
This seem breaking the gin plugin test

wu-sheng:
Which tool are you using to run shell on the windows?

mrproliu:
> > Do you use Windows for testing the plugin? I think maybe the `docker-compose` has a different name format for the container. For now, the plugins is works well in Linux and MacOS. I think you can judge the system environment and update the name dynamically.
> 
> 
> 
> I think we need to add a Windows e2e to verify at least one plugin? Is that possible? @mrproliu 

Sure. I can try it on tomorrow.

zheheBao:
> Which tool are you using to run shell on the windows?

I used the CentOS7 image to build a virtual machine in VMware
This is the result obtained by using the uname command in it

Vmvare

mrproliu:
Following this [question](https://stackoverflow.com/questions/69464001/docker-compose-container-name-use-dash-instead-of-underscore), I think you could check the version of the docker-compose.

zheheBao:
> Following this [question](https://stackoverflow.com/questions/69464001/docker-compose-container-name-use-dash-instead-of-underscore), I think you could check the version of the docker-compose.

Okay, I'll check it later

zheheBao:
> Following this [question](https://stackoverflow.com/questions/69464001/docker-compose-container-name-use-dash-instead-of-underscore), I think you could check the version of the docker-compose.

I am using a relatively new docker-compose version 2.17.3, which can be resolved by switching to 1.29.2 (the last version of 1.X). Therefore, I tried versions 2.0.0 and 2.0.1 again and found that the issue occurred again
So do we need to add a few lines of script to verify whether docker-compose 1.X or 2.X version, in order to give a prompt to newcomers who encounter my situation when running this script

mrproliu:
> So do we need to add a few lines of script to verify whether docker-compose 1.X or 2.X version, in order to give a prompt to newcomers who encounter my situation when running this script

I think you could add a version judgment and adapter between different docker-compose versions.

zheheBao:
> I think you could add a version judgment and adapter between different docker-compose versions.

Okay, I'll try it out later

wu-sheng:
So it's working on windows now?

zheheBao:
> So it's working on windows now?

Currently working on building Linux virtual machines on Windows

mrproliu (approved):
LGTM

PR #29, opened 2023-05-10
Split the build documentation into development version and user setup version

mrproliu:


docs/en/development-and-contribution/build-and-use-agent.md
```diff
@@ -0,0 +1,43 @@
+# Build and use the Agent from source codes
+
+When you want to build and use the Agent from source code, please follow these steps.
+
+## Install SkyWalking Go
+
+Use `go get` to import the latest version of `skywalking-go` program.
+
+```shell
+go get github.com/apache/skywalking-go@latest
```
wu-sheng, line 10:
```suggestion
// latest or any commit ID
go get github.com/apache/skywalking-go@latest
```

docs/en/development-and-contribution/build-and-use-agent.md
```diff
@@ -0,0 +1,43 @@
...
+```go
+import _ "github.com/apache/skywalking-go"
+```
+
+## Build the Agent
+
+When building the project, you need to clone the project and build it.
+
+```shell
+git clone https://github.com/apache/skywalking-go.git
```
wu-sheng, line 25:
```suggestion
// git clone the same version(tag or commit ID) as your dependency version.
git clone https://github.com/apache/skywalking-go.git
```

docs/en/development-and-contribution/build-and-use-agent.md
mrproliu: updated, I think we need add a comment when build the agent for make sure they using the same version(commit ID).

wu-sheng (approved):

PR #30, opened 2023-05-10
Skip tracing if gRPC connection is down

mrproliu:


wu-sheng (approved):

wu-sheng:
GHA seems down for a while.

PR #31, opened 2023-05-12
Add native http server plugin and tests

mrproliu:
1. Support monitor HTTP server in native HTTP plugin
2. Add HTTP plugin test
3. Update `Invocation` as an interface for plugin development, add the `ChangeArg` method to change the real method argument

### DIFF
--- a/docs/en/development-and-contribution/development-guide.md
+++ b/docs/en/development-and-contribution/development-guide.md
@@ -192,7 +192,7 @@ type Invocation interface {
 
     // IsContinue is the flag to control the method invocation, if it is true, the target method would not be invoked.
     IsContinue() bool
-    // DefineReturnValues is defined the return value of method, and continue the method invoked
+    // DefineReturnValues are defined the return value of the method, and continue the method invoked
     DefineReturnValues(...interface{})
 
     // SetContext is the customized context of the method invocation, it should be propagated the tracing span.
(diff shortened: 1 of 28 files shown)


docs/en/development-and-contribution/development-guide.md
```diff
@@ -172,29 +172,33 @@ The interceptor definition is as follows, you need to create a new structure and
...
+    // Args is get the arguments of the method, please cast to the specific type to get more information.
+    Args() []interface{}
+
+    // ChangeArg is change the argument value of the method
+    ChangeArg(int, interface{})
+
+    // IsContinue is the flag to control the method invocation, if it is true, the target method would not be invoked.
+    IsContinue() bool
+    // DefineReturnValues is defined the return value of method, and continue the method invoked
+    DefineReturnValues(...interface{})
```
wu-sheng, line 196: I want to confirm, this still works in the afterMethod, right?

wu-sheng:
```suggestion
    // DefineReturnValues are defined the return value of the method, and continue the method invoked
    DefineReturnValues(...interface{})
```

Please update comments of the method as well.

mrproliu: For now, It only works in the `BeforeMethod`.

wu-sheng: I think that is a little tricky, could you make it working in after as well? Through another PR of course.

mrproliu: Sure, I could update in other PR.

wu-sheng (approved):

wu-sheng:
Oops, I think we need to update the supported list?

mrproliu:
The docs have been updated and I have make the `DefineReturnValues` works in the `AfterMethod`.

wu-sheng (approved):

PR #33, opened 2023-05-15
Support dubbo plugin

mrproliu:
1. Support [dubbo-go](https://github.com/apache/dubbo-go) plugin
2. Remove unnecessary imports in **go.mod**


wu-sheng:
You seems to miss doc update again.

wu-sheng (approved):

wu-sheng:
Please add `CI / Required (pull_request)` depends on all plugin tests. Right now, the PR is able to merge before plugin tests passed.

plugins/dubbo/filter/graceful_shutdown/server_intercepter.go
```diff
@@ -0,0 +1,58 @@
...
+	url := filterInvoker.GetURL()
+	if url == nil {
+		return nil
+	}
+	s, err := tracing.CreateEntrySpan(generateOperationName(filterInvoker, dubboInv), func(k string) (string, error) {
+		attachment, _ := dubboInv.GetAttachment(k)
+		return attachment, nil
+	}, tracing.WithLayer(tracing.SpanLayerRPCFramework),
+		tracing.WithTag(tracing.TagURL, url.String()),
+		tracing.WithComponent(3))
```
wu-sheng, line 43: About this ID, should we consider some management/registration mechanism in the go? I know these are registered at the backend, at least, is this registration mentioned in go plugin dev doc?

mrproliu: I could add a section to help developers find the component ID. 
I think we don't need the management of the component ID registration in the go agent.

wu-sheng (approved):

PR #39, opened 2023-05-15
Support plugin exclusion mechanism

mrproliu:


docs/en/advanced-features/plugin-exclusion.md
```diff
@@ -0,0 +1,11 @@
+# Plugin Exclusion
+
+The plugin exclusion is used during the **compilation phase** to prevent certain plugins, as specified by their names,
```
wu-sheng, line 3:
```suggestion
The plugin exclusion is used during the **compilation phase** to exclude specific plugins, through their names, 
```

docs/en/advanced-features/plugin-exclusion.md
```diff
@@ -0,0 +1,11 @@
+# Plugin Exclusion
+
+The plugin exclusion is used during the **compilation phase** to prevent certain plugins, as specified by their names, 
+from being included in the compilation. Consequently, these excluded plugins will not generate corresponding data when the program is running, such as **Tracing**.
```
wu-sheng, line 4:
```suggestion
The plugin exclusion is used during the **compilation phase** to exclude specific plugins, through their names.
Consequently, the codes of these excluded plugins will not be weaved in, then, no relative tracing and metrics.
```

docs/en/advanced-features/plugin-exclusion.md
```diff
@@ -0,0 +1,19 @@
...
+Consequently, the codes of these excluded plugins will not be weaved in, then, no relative tracing and metrics.
+
+## Configuration
+
+
+```yaml
+plugin:
+  # List the names of excluded plugins, multiple plugin names should be splitted by ","
+  # NOTE: This parameter only takes effect during the compilation phase.
+  excluded: ${SW_AGENT_PLUGIN_EXCLUDES:}
```
wu-sheng, line 13: @liuhaoyang I added this, and update the config name(excludes -> excluded), as well as comments. Please update code accordingly.

mrproliu: fixed.

wu-sheng:
Should we have a plugin test for this? Such as Gin + Dubbo case, but only Gin spans are detected? Could you verify if is that possible and verified correctly?
About `verified correctly`, I mean, if excluded mechanism doesn't work properly, the gin + dubbo segment could trigger the test failing.

mrproliu:
Sure, I could add a plugin test for the plugin exclusion mechanism. HTTP and Gin as one test case, if ignore the HTTP framework, It should only have the Gin(server) (entry)span.

test/plugins/scenarios/plugin_exclusion/config/excepted.yml
```diff
@@ -0,0 +1,73 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+segmentItems:
+  - serviceName: plugin_exclusion
+    segmentSize: ge 3
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: GET:/health
```
wu-sheng, line 23: I think usually we don't check `/health` in Java practice.

test/plugins/scenarios/plugin_exclusion/config/excepted.yml
```diff
@@ -0,0 +1,73 @@
...
+            skipAnalysis: false
+            tags:
+              - {key: http.method, value: GET}
+              - {key: url, value: 'service:8080/health'}
+              - {key: status_code, value: '200'}
+      - segmentId: not null
+        spans:
+          - operationName: GET:/provider
+            parentSpanId: -1
+            spanId: 0
```
wu-sheng, line 42: I think we should indicate `ref` is null here to avoid the client doesn't propagate anything?

test/plugins/scenarios/plugin_exclusion/config/excepted.yml
mrproliu: I have deleted it in all plugin test cases.

test/plugins/scenarios/plugin_exclusion/config/excepted.yml
mrproliu: updated.

wu-sheng (approved):

PR #40, opened 2023-05-18
fix custome file do not take effect (#10821)

ethan256:


mrproliu (approved):
LGTM. Thanks. 
You reminded me if the user defines the customized configuration file without the environment key, we should remove the key in the default config. I will update it in another PR.

PR #41, opened 2023-05-19
Support go-restful framework with v3 version

mrproliu:
1. Support go-restful framework with v3 version
2. Support overwide operation names avoid to create multiple entry span
3. Support overwrite span tag

### DIFF
--- a/plugins/core/span_tracing.go
+++ b/plugins/core/span_tracing.go
@@ -28,11 +28,6 @@ import (
 )
 
 func NewSegmentSpan(defaultSpan *DefaultSpan, parentSpan SegmentSpan) (s SegmentSpan, err error) {
-	// if current and parent span are both entry span, then use parent span as segment span
-	if defaultSpan.SpanType == SpanTypeEntry && parentSpan != nil && parentSpan.IsEntry() {
-		parentSpan.SetOperationName(defaultSpan.OperationName)
-		return parentSpan, nil
-	}
 	ssi := &SegmentSpanImpl{
 		DefaultSpan: *defaultSpan,
 	}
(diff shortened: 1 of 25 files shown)


wu-sheng:
> Support overwide operation names to create multiple entry span

`to create multiple` or `avoid to create multiple`?

mrproliu:
> > Support overwide operation names to create multiple entry span
> 
> `to create multiple` or `avoid to create multiple`?

is `avoid to create multiple`, I have updated the description.

plugins/core/span_tracing.go
```diff
@@ -28,6 +28,11 @@ import (
 )
 
 func NewSegmentSpan(defaultSpan *DefaultSpan, parentSpan SegmentSpan) (s SegmentSpan, err error) {
+	// if current and parent span are both entry span, then use parent span as segment span
+	if defaultSpan.SpanType == SpanTypeEntry && parentSpan != nil && parentSpan.IsEntry() {
```
wu-sheng, line 32: Should we check parent span first before the span creation? It seems to cause a little extra payload?

mrproliu: I have advanced the parent check, please re-check

plugins/core/tracing.go
```diff
@@ -47,6 +47,11 @@ func (t *Tracer) CreateEntrySpan(operationName string, extractor interface{}, op
 	defer func() {
 		saveSpanToActiveIfNotError(ctx, s, err)
 	}()
+	// if parent span is entry span, then use parent span as result
+	if tracingSpan != nil && tracingSpan.IsEntry() {
+		tracingSpan.SetOperationName(operationName)
+		return tracingSpan, nil
+	}
```
wu-sheng, line 54: Do we have exit span overlap mechanism? In Java agent, the entry span is always using the newest created, but exit span usually uses the first time set.

wu-sheng (approved):
LGTM

wu-sheng (approved):
LGTM

plugins/core/tracing.go
mrproliu: For now, I haven't added the exit span overlap mechanism, I can add it in other PR.

PR #45, opened 2023-05-24
Refactor the global ID generator

mrproliu:
Following https://github.com/apache/skywalking/issues/10849, I have refactored the ID generator. 
ID generate format(same with java): `{instanceID}.{goid}.{timestamp}{sequence}`. We bind the `IDContext` to each goroutine(using TLS) to reduce the global locker.

Here are simple batch marks for comparison with UUID：
```
goos: darwin
goarch: amd64
pkg: github.com/apache/skywalking-go/plugins/core
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkGenerateID
BenchmarkGenerateID-16             	  956929	      1168 ns/op
BenchmarkGenerateIDParallels
BenchmarkGenerateIDParallels-16    	 4309476	       323.3 ns/op
BenchmarkUUID
BenchmarkUUID-16                   	 1000000	      1008 ns/op
BenchmarkUUIDParallels
BenchmarkUUIDParallels-16          	  856684	      1341 ns/op
PASS
```

I will do another bench mark for the agent, If there is no performance issue, I will remove the `UUID` dependency.

### DIFF
--- a/plugins/core/context.go
+++ b/plugins/core/context.go
@@ -33,6 +33,7 @@ type ContextSnapshoter interface {
 type TracingContext struct {
 	activeSpan TracingSpan
 	Runtime    *RuntimeContext
+	ID         *IDContext
 }
 
 func (t *TracingContext) TakeSnapShot(val interface{}) interface{} {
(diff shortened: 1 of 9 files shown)


wu-sheng:
Please indicate how much is the improvement?

mrproliu:
In the UUID default implementation, they use the global locker when generate. In the current implementation, we haven't any locker. 
Based on the benchmark result (parallel), it will be about 3-4 times faster than `UUID`.

wu-sheng (approved):

PR #46, opened 2023-05-27
Support go-micro framework with v4 version

mrproliu:


test/plugins/scenarios/microv4/config/excepted.yml
```diff
@@ -0,0 +1,85 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+segmentItems:
+  - serviceName: micro-server
+    segmentSize: ge 1
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: go.micro.srv.greeter.Say.Hello
```
wu-sheng, line 23: I think the only concern of this implementation would be, this would never be an `endpoint`, and not in the `endpoint dependency` chain as well.

wu-sheng: In this particular case, I think it falls into one scenario we discussed. Whether span should a wrapper to update cross-goroutine. Such as here, should we consider no microv4 span, but HTTP server span would be updated later.

docs/en/development-and-contribution/development-guide.md
```diff
@@ -387,6 +387,26 @@ Following the previous API define, you should following these steps to use the a
...
 5. When the `span.AsyncFinish()` is complete for all spans, the all spans would be finished and report to the backend.
 
+#### Tracing Context Operation
+
+In the Go Agent, by default, passes a snapshot of the current Tracing Context when crossing goroutines. 
+However, if you prefer manual management of the Tracing Context across multiple goroutines, please use the following APIs.
+
+```go
+// CaptureContext capture current tracing context in the current goroutine.
+func CaptureContext()
```
wu-sheng, line 397: Missing return value?

docs/en/development-and-contribution/development-guide.md
```diff
@@ -387,6 +387,26 @@ Following the previous API define, you should following these steps to use the a
 4. Once the above steps are all set, call `span.AsyncFinish()` in any goroutine.
 5. When the `span.AsyncFinish()` is complete for all spans, the all spans would be finished and report to the backend.
 
+#### Tracing Context Operation
+
+In the Go Agent, by default, passes a snapshot of the current Tracing Context when crossing goroutines. 
+However, if you prefer manual management of the Tracing Context across multiple goroutines, please use the following APIs.
```
wu-sheng, line 393:
```suggestion
In the Go Agent, Trace Context would continue cross goroutines automatically by default.
However, in some cases, goroutine would be context sharing due to be scheduled by the pool mechanism. Consider these advanced APIs to manipulate context and switch the current context.
```

docs/en/development-and-contribution/development-guide.md
```diff
@@ -387,6 +387,26 @@ Following the previous API define, you should following these steps to use the a
...
+func CaptureContext()
+
+// ContinueContext continue the tracing context in the current goroutine.
+func ContinueContext(ctx ContextSnapshot)
+
+// CleanContext clean the tracing context in the current goroutine.
+func CleanContext()
+```
+
+Following the previous API define, you should following these steps to use the tracing context API:
```
wu-sheng, line 406:
```suggestion
Typically, use APIs as following to control or switch the context:
```

docs/en/development-and-contribution/development-guide.md
mrproliu: updated.

wu-sheng (approved):

wu-sheng:
It seems testing fails. Please fix.

tools/go-agent/instrument/plugins/rewrite/var.go
```diff
@@ -19,15 +19,14 @@ package rewrite
...
-	"strings"
 
 	"github.com/dave/dst"
 )
 
 func (c *Context) Var(val *dst.ValueSpec, onlyName bool) {
 	if len(val.Names) == 1 {
 		oldName := val.Names[0].Name
-		if !strings.HasPrefix(oldName, GenerateCommonPrefix) {
+		if !c.alreadyGenerated(oldName) {
```
wu-sheng, line 29: Something could be rewritten twice? Could you explain the case?

mrproliu: Yes, I could rewrite it twice. 
1. Generate name mapping relationships for all top-level types, which will be used by all Go files that reference them to quickly find mappings based on names. At this point, it is used only for name changes.
2. Once the top-level types are generated, all content will be rewritten. This method may be executed again, and will also be re-executed when variables are defined within the method.

wu-sheng (approved):

PR #55, opened 2023-06-06
Add support for go-redis

jiekun:
# Change
## Added
- Added support for [go-redis v9](https://github.com/redis/go-redis).

## Note
- [x] document is updated if necessary.
- [x] New features are tested.


wu-sheng:
The test is also expected. Pleaser read the doc.

wu-sheng:
You should be able to run this test and make sure it passed, rather than relying on the limited upstream resource to keep rerunning.

jiekun:
> You should be able to run this test and make sure it passed locally, rather than relying on the limited upstream resource to keep rerunning.

Just figure out how to run it locally. 

The network inside GFW is poor. When I run the integration test, some packages are not able to be downloaded, which needs a lot of change in the .tpl file to use a proxy.

```
// so those are generated according to the .tpl. System proxy does not work inside the docker container 
// so I have to try it step by step.....
RUN go env -w GO111MODULE=on
RUN go env -w GOPROXY=https://proxy.golang.com.cn,direct
RUN go mod tidy
```

Sorry for pushing my commit frequently.

jiekun:
> This is always better to run on a cloud ubuntu to make all easier. Let me know if you need any help or have passed locally.

Thank you sir. The test workflow is fine on my local PC now. There are some issues to be fixed tho.

BTW I would like to say the documentation is very clean, thanks to @mrproliu . I'd like to write a new blog post about *how to extend new plugin to skywalking-go* with some detail example. Actually translating those doc into Chinese is already **good enough**, but I would like to have a **step by step demonstration** to help more developers who's new to SkyWalking :)

wu-sheng:
You are welcome to add this to our website, CN blog channel. And we could forward this to [juejin](https://juejin.cn/user/13673577331607/posts) as well.

jiekun:
Thanks.

The compilation is failing with error as below. I think most of my codes are following the development guide. Spent an hours to figure it out but still cannot locate the root cause.
```
/usr/local/go/bin/go build -toolexec /home/jiekun/repo/github.com/skywalking-go/test/plugins/dist/skywalking-go-agent -debug /home/jiekun/repo/github.com/skywalking-go -a -o /home/jiekun/.cache/JetBrains/GoLand2023.1/tmp/GoLand/___10go_build_main_go -gcflags all=-N -l /home/jiekun/repo/github.com/skywalking-go/test/plugins/scenarios/go-redisv9/main.go #gosetup
# github.com/redis/go-redis/v9
/home/jiekun/repo/github.com/skywalking-go/plugins/go-redisv9/hook.go:157: undefined: redis
/home/jiekun/repo/github.com/skywalking-go/plugins/go-redisv9/intercepter.go:44: undefined: redis
/home/jiekun/repo/github.com/skywalking-go/plugins/go-redisv9/intercepter.go:46: undefined: redis
/home/jiekun/repo/github.com/skywalking-go/plugins/go-redisv9/intercepter.go:52: undefined: redis
2023/06/06 22:00:59 exit status 2
```

I would like to have a few *hints* from maintainers / reviewers if possible (orz), and continue on this enhencement tomorrow.

wu-sheng:
@mrproliu When you have time.

mrproliu:
@jiekun You need to add the `-work` parameter and remove the `-debug` parameter temporarily in the build command, then you can get the real generate files in the console. Please deep into the file you are compiling error, and see what's happening in there.

jiekun:
> You need to add the `-work` parameter and see what's happening in there.

Thanks. I did it yesterday and I am going to post the source file and generated file here as below:
```go
        // source

	switch c := rdb.(type) {
	case *redis.Client:
		c.AddHook(newRedisHook(c.Options().Addr))
	case *redis.ClusterClient:
		c.AddHook(newRedisHook(""))

		c.OnNewNode(func(rdb *redis.Client) {
			rdb.AddHook(newRedisHook(rdb.Options().Addr))
		})
	case *redis.Ring:
		c.AddHook(newRedisHook(""))

		c.OnNewNode(func(rdb *redis.Client) {
			rdb.AddHook(newRedisHook(rdb.Options().Addr))
		})
```

```go
        // generated

	switch c := rdb.(type) {
	case *redis.Client:
		c.AddHook(skywalking_operatorStaticMethodRedisnewRedisHook(c.Options().Addr))
	case *redis.ClusterClient:
		c.AddHook(skywalking_operatorStaticMethodRedisnewRedisHook(""))

		c.OnNewNode(func(rdb *Client) {
			rdb.AddHook(skywalking_operatorStaticMethodRedisnewRedisHook(rdb.Options().Addr))
		})
	case *redis.Ring:
		c.AddHook(skywalking_operatorStaticMethodRedisnewRedisHook(""))

		c.OnNewNode(func(rdb *Client) {
			rdb.AddHook(skywalking_operatorStaticMethodRedisnewRedisHook(rdb.Options().Addr))
		})
```

The `*redis` in `switch-case` line is not replaced. Initially I thought it's because those structure are not instrumented. However same struct in function call is replaced (e.g. `c.OnNewNode(func(rdb *redis.Client) {` is replaced to `c.OnNewNode(func(rdb *Client) {`). Any clue @mrproliu ?

mrproliu:
Ok, that's related to the [rewrite processing](https://github.com/apache/skywalking-go/tree/main/tools/go-agent/instrument/plugins/rewrite). At present, it may not cover all the scenarios that need to be rewritten. As the version iterations, the function will become more and more perfect. You can have a debug with the following steps:
1. Find the arguments of compile: The `instrument.log` file can be found in the **compiled temporary directory**, where you can find the parameter information of your current compiled package.
2. Debug the agent tools: Debug [Agent Tool](https://github.com/apache/skywalking-go/tree/main/tools/go-agent/cmd) with the arguments. 

If it's not easy for you, I can give you some help.

jiekun:
Well I have little experience with code generation and ast parsing. I think I would prefer some help here (orz). Not sure how it could perform tho, like rebasing a new branch with the rewritten support for `switch-case`?

I would like to provide my guess of the missing supports:
1. `switch-case` as addressed above: https://github.com/apache/skywalking-go/pull/55/files#diff-67660820b0f7f0cf5ec1b1c22b073b5005a83ac5112535be7b54863f2f4d332dR43
2. It looks like `const` (`redis.Nil`) is not rewritten as well: https://github.com/apache/skywalking-go/pull/55/files#diff-9bfc186e1c784e2abf6fe20c851745fcc16d1f4719051e1ff95e077047245445R163

mrproliu:
@jiekun The rewrite-related problem has been fixed, please re-check locally.

jiekun:
@mrproliu Thank you for giving me a hand with the rewrite function.
@wu-sheng I'm removing the draft status. Please review it when you have time. Thank you all.

test/plugins/scenarios/go-redisv9/config/excepted.yml
```diff
@@ -0,0 +1,93 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+segmentItems:
+  - serviceName: go-redisv9
+    segmentSize: ge 1
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: set
```
mrproliu, line 23: Please add "redis/" prefix for each operation name to let the user easy to know is a Redis operation.

test/plugins/scenarios/go-redisv9/config/excepted.yml
```diff
@@ -0,0 +1,93 @@
...
+            endTime: nq 0
+            componentId: 5014
+            isError: false
+            spanType: Exit
+            peer: redis-server:6379
+            skipAnalysis: false
+            tags:
+              - {key: cache.type, value: redis}
+              - { key: cache.args, value: not null}
+          - operationName: redis.pipeline set set
```
mrproliu, line 51: don't add a blank in an operation name, split by "/" is a good choice.

mrproliu (commented):
Please add the plugin to the documentation and fix the comments.

test/plugins/scenarios/go-redisv9/config/excepted.yml
jiekun: So the pipeline operation will shows up like `redis/pipeline/set/set/get/hmset/...`. Wondering if we could have a better format for op name with multiple commands.

I'm changing to this format for now. But feel free to comment with other idea before we merge this into master.

wu-sheng: This would be a good refer, https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/jedis-4.x-scenario/config/expectedData.yaml

After all, Java plugins are existing for a long time.

wu-sheng:
Docs should be updated accordingly, https://github.com/apache/skywalking-go/blob/main/docs/en/agent/support-plugins.md

test/plugins/scenarios/go-redisv9/config/excepted.yml
wu-sheng: Batch is tricky actually, and easy to cause overload.

If I read the codes correctly, https://github.com/apache/skywalking-java/blob/6621f82739340bbaaac2338cb06ccbe3d38513a5/apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/JedisMethodInterceptor.java#L31

test/plugins/scenarios/go-redisv9/config/excepted.yml
```diff
@@ -0,0 +1,93 @@
...
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 5014
+            isError: false
+            spanType: Exit
+            peer: redis-server:6379
+            skipAnalysis: false
+            tags:
+              - {key: cache.type, value: redis}
+              - {key: cache.args, value: not null}
```
wu-sheng, line 36: `cache.op` tag seems not being added.

Please refer to virtual cache docs, https://skywalking.apache.org/docs/main/next/en/setup/service-agent/virtual-cache/

Those tags are important if you expect all telemetry data is available on the dashboard.

The catalog we are using at Java side could be found here

https://github.com/apache/skywalking-java/blob/6621f82739340bbaaac2338cb06ccbe3d38513a5/apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/JedisPluginConfig.java

test/plugins/scenarios/go-redisv9/config/excepted.yml
jiekun: well, based on discussion above I am trying to proposal span like this:

```
  - operationName: redis/pipeline  # could we just use pipeline directly, it's clear in "redis context"
    parentSpanId: 0
    spanId: 4
    spanLayer: Cache
    startTime: nq 0
    endTime: nq 0
    componentId: 5014
    isError: false
    spanType: Exit
    peer: redis-server:6379
    skipAnalysis: false
    tags:
      - {key: cache.type, value: redis}
      - {key: cache.cmd, value: 'pipeline:get/get'}  # retain necessary information for debugging
```

wu-sheng: `pipeline:get/get` Why `get/get`?

test/plugins/scenarios/go-redisv9/config/excepted.yml
jiekun: Tags are added accordingly.

test/plugins/scenarios/go-redisv9/config/excepted.yml
jiekun: > `pipeline:get/get` Why `get/get`?

because two different command are executed with `pipeline`. In this test case I execute the following codes:
```
# start pipeline
GET xxxx
GET yyyy
# submit my pipline for execution
```

Other example of the tag value:
```
pipeline:set/get/get/get/set

# when len(op) > 10
pipeline:hmset/get/set/get/get/get/get/get/get/get/...
```

mrproliu (approved):
General LGTM.

PR #56, opened 2023-06-07
Support running e2e tests on MacOS

ethan256:


wu-sheng:
Should we add a new CI env of MacOS to verify this?

mrproliu:
> Should we add a new CI env of MacOS to verify this?

Yes. @ethan256 Please add a simple GHA to verify the plugin test on MacOS. a single plugin testing is enough, such as the native HTTP plugin.

test/plugins/run.sh
```diff
@@ -155,16 +155,24 @@ for framework_version in $frameworks; do
...
-  sed -i "s/^go [0-9]*\.[0-9]*/go ${go_version}/" go.mod
   # append the module name(for go1.20, module name cannot be same in same workspace)
   mod_case_name=$(echo "${case_name}" | sed 's/\//_/g; s/\./_/g; s/-/_/g')
   mod_name=$(head -n 1 go.mod | cut -d " " -f 2)
-  sed -i "s/^module /module ${mod_case_name}\//" go.mod
-  find . -name "*.go" -type f -exec sed -i "s|${mod_name}|${mod_case_name}/${mod_name}|g" {} \;
-
-  # ajust the plugin replace path
-  sed -i -E '/^replace/ s#(\.\./)#\1../#' go.mod
+  if [ $(uname) = 'Darwin' ]; then
```
lujiajing1126, line 161: I would suggest moving conditional check to the top of the file, for example with 'alias' or creating a new function?

lujiajing1126 (changes requested):

test/plugins/run.sh
mrproliu: I think a new function is more reasonable. @ethan256 what do you think?

.github/workflows/plugin-tests.yaml
```diff
@@ -85,18 +98,58 @@ jobs:
...
       - test
+      - test-on-macos
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
       - name: Merge Requirement
         run: |
           if [[ ${{ needs.test.result }} != 'success' ]]; then
             exit -1
-          fi
```
mrproliu, line 102: Please add the MacOs test result checker.

lujiajing1126 (approved):
LGTM

wu-sheng:
@lujiajing1126 Please make sure you have commit logs reset when squash. It is very crazy to see this on commit log.

image

Most are only noises.

lujiajing1126:
> @lujiajing1126 Please make sure you have commit logs reset when squash. It is very crazy to see this on commit log.
> 
> image
> Most are only noises.

I will notice next time

PR #59, opened 2023-06-13
feature: Add support for Kafka client sarama

jiekun:
### Feature
- Added support for [sarama](https://github.com/Shopify/sarama).

### Note
- [ ] document is updated if necessary. (Not yet)
- [ ] New features are tested. (Not yet)


jiekun:
I am posting a draft here. 

I got the following test result:
```yaml
segmentItems:
- serviceName: sarama
  segmentSize: 3
  segments:
  - segmentId: e6ddbe3309dc11ee98fe0242ac130005.42.38984090918090002
    spans:
    - operationName: GET:/health
      parentSpanId: -1
      spanId: 0
      spanLayer: Http
      startTime: 1686655491809
      endTime: 1686655491809
      componentId: 5004
      isError: false
      spanType: Entry
      peer: ''
      skipAnalysis: false
      tags:
      - {key: http.method, value: GET}
      - {key: url, value: 'service:8080/health'}
      - {key: status_code, value: '200'}
  - segmentId: e6ddbe3309dc11ee98fe0242ac130005.14.38984090918390002
    spans:
    - operationName: Kafka/sarama_auto_instrument/Producer
      parentSpanId: -1
      spanId: 0
      spanLayer: MQ
      startTime: 1686655491839
      endTime: 1686655491839
      componentId: 5015
      isError: false
      spanType: Exit
      peer: kafka-server:9092
      skipAnalysis: false
      tags:
      - {key: mq.broker, value: 'kafka-server:9092'}
      - {key: mq.topic, value: sarama_auto_instrument}
  - segmentId: e6ddbe3309dc11ee98fe0242ac130005.62.38984090920720001
    spans:
    - operationName: Kafka/sarama_auto_instrument/Consumer
      parentSpanId: -1
      spanId: 0
      spanLayer: MQ
      startTime: 1686655492072
      endTime: 1686655492072
      componentId: 5015
      isError: false
      spanType: Entry
      peer: ''
      skipAnalysis: false
      tags:
      - {key: mq.broker, value: 'kafka-server:9092'}
      - {key: mq.topic, value: sarama_auto_instrument}
      refs:
      - {parentEndpoint: Kafka/sarama_auto_instrument/Producer, networkAddress: 'kafka-server:9092',
        refType: CrossProcess, parentSpanId: 0, parentTraceSegmentId: e6ddbe3309dc11ee98fe0242ac130005.14.38984090918390002,
        parentServiceInstance: e6dde1da09dc11ee98fe0242ac130005@172.19.0.5, parentService: sarama,
        traceId: e6ddbe3309dc11ee98fe0242ac130005.14.38984090918390001}
meterItems: []
logItems: []
```

It seems an HTTP server span is missing. The workflow shoule be:
```
Trace 1: [HTTP: /health]
Trace 2: [HTTP: /execute] -> [Kafka Producer] -> [Kafka Consumer]
```

So that we should get 2 trace and 4 span here. It seems the `[HTTP /execute]` is missing.

I would like to ask @mrproliu about how the context is passing through in-process function call (without `ctx context.Context`), not sure if it's relevent. I assume we should get `[Kafka Producer]` span with `parentSpandId: 0` (which should be the span id of HTTP server)

jiekun:
Ok I will check what happen to the HTTP interceptor.

I am going to fix an small issue when instrumenting new plugin:
`tools/go-agent/instrument/plugins/instrument.go`

Existing codes does not support package with upper-case character.
`github.com/Shopify/sarama` -> `github.com/!shopify/sarama` and cannot be identified properly.

Please review this change as well.

jiekun:
> > Existing codes does not support package with upper-case character.
> 
> When the package name would be `github.com/!shopify/sarama`?

Yes. I got the variable value by printing and debugging. Following the official [document](https://pkg.go.dev/golang.org/x/mod/module#hdr-Escaped_Paths) I assume it's caused by path escape.

I'm not 100% sure about it and hence request for a proper review.

jiekun:
> I think this is not a good practice to keep two discussions in PR. Please open a new one about what you proposed to change for the kernel level. And for that, provide details of hierarchy structures of projects or a re-produrable env.

Just raised a new PR: https://github.com/apache/skywalking-go/pull/64 for this issue

jiekun:
@mrproliu Sorry for bothering ya again.

Sarama provided `producer & consumer interceptor` (interface as below), but those function are called by worker goroutine (which communicate with user goroutine via `chan`). So it seems that the span context in user goroutine cannot be aware by worker goroutine.

Any idea to keep those instrument inside the border of `producer & consumer interceptor`?

```go
// ProducerInterceptor allows you to intercept (and possibly mutate) the records
// received by the producer before they are published to the Kafka cluster.
// https://cwiki.apache.org/confluence/display/KAFKA/KIP-42%3A+Add+Producer+and+Consumer+Interceptors#KIP42:AddProducerandConsumerInterceptors-Motivation
type ProducerInterceptor interface {

	// OnSend is called when the producer message is intercepted. Please avoid
	// modifying the message until it's safe to do so, as this is _not_ a copy
	// of the message.
	OnSend(*ProducerMessage)
}

// ConsumerInterceptor allows you to intercept (and possibly mutate) the records
// received by the consumer before they are sent to the messages channel.
// https://cwiki.apache.org/confluence/display/KAFKA/KIP-42%3A+Add+Producer+and+Consumer+Interceptors#KIP42:AddProducerandConsumerInterceptors-Motivation
type ConsumerInterceptor interface {

	// OnConsume is called when the consumed message is intercepted. Please
	// avoid modifying the message until it's safe to do so, as this is _not_ a
	// copy of the message.
	OnConsume(*ConsumerMessage)
}
```

I tried the following way:
1. intercept the user goroutine `SendMessage(msg *ProducerMessage) (partition int32, offset int64, err error)` method, inject trace context to `msg.Header` (a Key-Value struct).
2. (msg is sent to worker goroutine via `chan`).
3. In `OnSend(*ProducerMessage)`, retrive the trace context from `msg.Header` and try to generate a new exit span base on it.

issues:
1. `SyncProducer` has `SendMessage` method, but `AsyncProducer` only provide a `chan` for user to feed msg to the worker (`chan<-msg`). So there is no method similar to `SendMessage(msg *ProducerMessage)` could be intercept.
2. well it seems we could not **retrive** a trace context stored in somewhere, which is active in another goroutine?

I would like to have your opinion about how we do instrument on `chan` related scenario. Maybe I just miss some available components, appreciated if you could share your idea :)

mrproliu:
> Sarama provided producer & consumer interceptor (interface as below), but those function are called by worker goroutine (which communicate with user goroutine via chan).

As you mentioned, I think the better way is to create an interceptor on the user goroutine(which should same as the HTTP handler goroutine to encase the HTTP has already been finished). 

> In OnSend(*ProducerMessage), retrive the trace context from msg.Header and try to generate a new exit span base on it.

On the worker goroutine(which is now intercepted I think?), you don't need to do anything because you have already added the tracing context header for the message. It can receive the tracing context on the server side maybe?

> SyncProducer has SendMessage method, but AsyncProducer only provide a chan for user to feed msg to the worker (chan<-msg). So there is no method similar to SendMessage(msg *ProducerMessage) could be intercept.

I'm not sure if I understand what you mean. For the async mode, I think you should also add the tracing context header in the message. If the message which needs to add to the channel is not the original Kafka message, you can [enhance the message from official documentation](https://skywalking.apache.org/docs/skywalking-go/next/en/development-and-contribution/development-guide/#struct-matcher) and add the tracing context header to it. Then, you should intercept the message reading from the channel, and add the tracing context from the enhanced instance into the real Kafka message.

wu-sheng:
We usually manipulate the metada/header of message to propagate the context if that is your asking.

jiekun:
The `message` struct in golang/sarama pkg provide a `Header` field for us to set something.

I am going to provide more context of the current situation here

## SyncProducer
SyncProducer provide a `SendMessage` method. User just call `SendMessage(msg)` to send it to kafka.

```
func SendMessage(msg) {
    workerChannel <- msg
    ...
}
```

So after instrumenting, this method should be like:
```go
func SendMessage(msg) {
    StartSpan()   // This op also inject trace ctx to the msg Header field

    workerChannel <- msg
    ...

    EndSpan()
}
```

Ok it's a normal instrumentation like what we did in other plugins.

## AsyncProducer
However, for the `AsyncProducer`, the basic usage is:
```
func main() {
    msg := &ProducerMessage{Topic: "my_topic", Value: StringEncoder(TestMessage)}
    producer.Input() <- msg
}
```
It's a bit complicated than the `SyncProducer` case. We need to add trace context **when instantiating the struct**.

According to the doc:
> Next, you can set custom content for the specified enhanced instance when intercepting methods.

We have to intercept a method before operating the dynamic field. But it seems we do not have method to be intercept since use throw the msg to the channel in `main`.

mrproliu:
I understand your concern regarding the `AsyncProducer`. Adding tracing context into the message is difficult as there is no method to intercept the instanced struct or hook method. However, there are a couple of ways to implement this:
1. User API: We can provide an API that allows users to manually propagate the tracing context while operating with the message.
2. Line-level instrument: We can offer a line-level instrument that ignores the package filter for injecting code into the user code. The plugin can then acknowledge when the struct has been instanced.

If you have any other ideas, please let me know. However, for now, we may need to accept that we cannot trace the async producer if we cannot intercept any methods.

jiekun:
OK Thanks for clarify it. I will note the behaviour of async producer in document as well.

wu-sheng:
> Line-level instrument: We can offer a line-level instrument that ignores the package filter for injecting code into the user code. The plugin can then acknowledge when the struct has been instanced.

This works, but very tricky. Because the line of codes are much easier to change, and make things unstable.

User APIs are the way in Java for now

mrproliu:
> > Line-level instrument: We can offer a line-level instrument that ignores the package filter for injecting code into the user code. The plugin can then acknowledge when the struct has been instanced.
> 
> This works, but very tricky. Because the line of codes are much easier to change, and make things unstable.

Yes, If we need this feature, It should limit what kind of statement could be enhanced. 

> User APIs are the way in Java for now

In this case, maybe we can add Span API for user creation manually. But will lose the real message sent (in another goroutine).

jiekun:
1 design to be review, and 1 issue remaining.

## To be review
The following implementation still needs review.

- As discussed, message sent with sync producer will carry the trace info in user goroutine.
- `OnSend` hook is still instrumented, which intercept **all** messages sent to kafka. Messages which missing trace info in message `Header` are considered **sent by async producer**. Trace info (without parent span, a.k.a [`spanId 0, parentId -1`]) will be appended. I consider it as a kind of **Best-effort delivery** pattern here.

```yaml
segmentItems:
  - serviceName: sarama
    segmentSize: ge 3
    segments:
      # Health Check Trace
      - segmentId: not null
        spans:
          - operationName: GET:/health
            parentSpanId: -1
            spanId: 0
            ...

      # Sync Producer Trace, HTTP span(spanId=0) -> Producer Span(spanId=1)
      - segmentId: not null
        spans:
          - operationName: Kafka/sarama_auto_instrument/Producer
            parentSpanId: 0
            spanId: 1
            ...
          - operationName: GET:/execute
            parentSpanId: -1
            spanId: 0
            ...

      # Consumer Trace for message via sync producer, ref parentSpanId=1
      - segmentId: not null
        spans:
          - operationName: Kafka/sarama_auto_instrument/Consumer
            parentSpanId: -1
            spanId: 0
            ...
            refs:
              - {parentEndpoint: 'GET:/execute', networkAddress: 'kafka-server:9092', refType: CrossProcess,
                 parentSpanId: 1, parentTraceSegmentId: not null,
                 parentServiceInstance: not null, parentService: sarama,
                 traceId: not null}

      # Async Producer Trace, parentSpanId = -1
      - segmentId: not null
        spans:
          - operationName: Kafka/sarama_auto_instrument/Producer
            parentSpanId: -1
            spanId: 0
            ...
      
      # Consumer Trace for message via async producer, ref parentSpanId=0
      - segmentId: not null
        spans:
          - operationName: Kafka/sarama_auto_instrument/Consumer
            parentSpanId: -1
            spanId: 0
            ...
            refs:
              - { parentEndpoint: 'Kafka/sarama_auto_instrument/Producer', networkAddress: 'kafka-server:9092', refType: CrossProcess,
                  parentSpanId: 0, parentTraceSegmentId: not null,
                  parentServiceInstance: not null, parentService: sarama,
                  traceId: not null}

meterItems: []
logItems: []
```

## Issue
Have trouble instrumenting a slice input:
```
                // SendMessage is instrumented correctly.
		{
			At: instrument.NewMethodEnhance(
				"*syncProducer",
				"SendMessage",
				instrument.WithArgsCount(1),
				instrument.WithArgType(0, "*ProducerMessage"),
				instrument.WithResultCount(3),
				instrument.WithResultType(0, "int32"),
				instrument.WithResultType(1, "int64"),
				instrument.WithResultType(2, "error"),
			),
			Interceptor: "SendMessageInterceptor",
		},
                // Attempt to instrument SendMessages but error happened.
		//{
		//	At: instrument.NewMethodEnhance(
		//		"*syncProducer",
		//		"SendMessages",
		//		instrument.WithArgsCount(1),
		//		instrument.WithArgType(0, "[]*ProducerMessage"),
		//		instrument.WithResultCount(1),
		//		instrument.WithResultType(0, "error"),
		//	),
		//	Interceptor: "SendMessagesInterceptor",
		//},
```

Logs:
```
# github.com/Shopify/sarama
/tmp/go-build3018337911/b025/skywalking_delegator.go:328:111: undefined: skywalking_operatorTypeSaramaProducerMessage
/tmp/go-build3018337911/b025/skywalking_delegator.go:342:21: undefined: skywalking_operatorTypeSaramaProducerMessage
```

tmp files:
```
// For SendMessage
func skywalking_enhance_github_com_Shopify_sarama_syncProducerSendMessage(recv_0 **syncProducer, param_0 **ProducerMessage) (ret_0 int32, ret_1 int64, ret_2 error, invocation *skywalking_operatorTypeOperatorrealInvocation, skip bool) {

// For SendMessages, which contains error
func skywalking_enhance_github_com_Shopify_sarama_syncProducerSendMessages(recv_0 **syncProducer, param_0 *[]*skywalking_operatorTypeSaramaProducerMessage) (ret_0 error, invocation *skywalking_operatorTypeOperatorrealInvocation, skip bool)
```

The data struct `*skywalking_operatorTypeSaramaProducerMessage` is not existed in generated files. 

Well I'm going to report this issue first to see if it's a bug. AFK utill Monday.

mrproliu:
> Messages which missing trace info in message Header are considered sent by async producer. 

We should **not** instrument if there is no tracing context we can get. Because It's not meaningful and will generate the wrong traffic(like CPM metrics) for the service/instance relation(sync mode already sends it, and `OnSend` send it again).

> The data struct `*skywalking_operatorTypeSaramaProducerMessage` is not existed in generated files.

It would help if you deep dived into it to see what's happening. I have given you [a solution](https://github.com/apache/skywalking-go/pull/55#issuecomment-1579737338) to this kind of problem before.

jiekun:
I'm still working on the file rewriting.

As suggested, I try to debug the `dstutil.Apply` func. It's still difficult because when I see those AST tree and rewrite, I have no idea which is right and which is wrong, and why it's rewritten into this bla bla bla.... Hence I am still not able to do the fix or find the cause.

I'm going to simplify the issue here, based on the clues I've found.

type `*ProducerMessage` is parsed into the following form:
<a href="https://imgbox.com/s9EpVkRN" target="_blank">image host</a>

type `[]*ProducerMessage` is parsed into the following form:
<a href="https://imgbox.com/1yEIWHvw" target="_blank">image host</a> 

After written, `*ProducerMessage` remain the same. And `[]*ProducerMessage` became `[]*skywalking_operatorTypeSaramaProducerMessage`.

And error happened: 
```
# github.com/Shopify/sarama
/tmp/go-build3018337911/b025/skywalking_delegator.go:328:111: undefined: skywalking_operatorTypeSaramaProducerMessage
/tmp/go-build3018337911/b025/skywalking_delegator.go:342:21: undefined: skywalking_operatorTypeSaramaProducerMessage
```

1. which is correct? should the array param be rewritten? should `*ProducerMessage` be rewritten?
- If the slice param needs to be rewritten, then I need to check why the tool did not generate a new structure.
- If the slice param should not be rewritten, that's a lot more harder since the AST code is complex...

orz. I still don't know if it's a bug or a feature, which block me from going further

mrproliu:
After examining your screenshot, it appears that the issue arises when the method parameters in the delegator are overridden.

For framework-related type overriding, the agent adds the current package prefix to the relevant data content when writing into the delegator (before overriding). This way, the current content is recognized as a framework object during overriding.

You can resolve this issue by [adding a case for ArrayType in this method](https://github.com/apache/skywalking-go/blob/main/tools/go-agent/tools/enhancement.go#L193-L224).

wu-sheng:
@jiekun There are two weeks, anything update?

PR #62, opened 2023-06-15
Add mongodb client plugin

mrproliu:


docs/en/agent/support-plugins.md
```diff
@@ -13,5 +13,6 @@
 * Database Client
   * `gorm`: [GORM](https://github.com/go-gorm/gorm)
     * [MySQL Driver](https://github.com/go-gorm/mysql)
+  * `mongo`: [Mongo](https://github.com/mongodb/mongo-go-driver)
```
wu-sheng, line 16: Please the whole support-plugins.md to add tested versions.

wu-sheng: I think we missed that before.

mrproliu: updated.

wu-sheng (approved):

PR #68, opened 2023-07-11
Support log report to the backend

mrproliu:
image
image
image


wu-sheng:
Let's update the docs for metrics and logs. We should re-organize the menu first.

wu-sheng (approved):

PR #69, opened 2023-07-12
Enhance the `logrus` logger plugin to support adapt without any settings method invoke

mrproliu:


docs/en/advanced-features/logging-setup.md
```diff
@@ -5,6 +5,16 @@ It currently supports the recognition of `Logrus` and `Zap` frameworks. If neith
...
 
+## Logging Detection
+
+Log detection means that the logging plugin would automatically detect the usage of logs in your application. 
+When the log type is set to `auto`, it would choose the appropriate log based on the creation rules of different frameworks. The selection rules vary depending on the framework:
+
+1. `Logrus`: It automatically selects the current logger when executing functions such as `logrus.New`, `logger.SetOutput`, or `logger.SetFormatter`.
+2. `Zap`: It automatically selects the current logger when executing functions such as `zap.New`, `zap.NewNop`, `zap.NewProduction`, `zap.NewDevelopment`, or `zap.NewExample`.
+
+If there are multiple different logging systems in your current application, the last-called logging system would be chosen.
```
wu-sheng, line 16: What does `last-called` mean? Multiple logging systems usually mean hybrid logging usage, right?

mrproliu: > Multiple logging systems usually mean hybrid logging usage, right?

Yes.

> What does `last-called` mean?

That means which logging system would be used on the agent logging. Hybrid logging usage permits logging uploader and tracing enhancement, but the Go Agent logging only needs to send the log to the one logging system.

wu-sheng: OK, got it.

wu-sheng (approved):

PR #78, opened 2023-07-19
Support enhance vendor management project

mrproliu:


wu-sheng (approved):

tools/go-agent/tools/vendor.go
```diff
@@ -0,0 +1,31 @@
...
+// under the License.
+
+package tools
+
+import "strings"
+
+const vendorDir = "/vendor/"
+
+// UnVendor removes the vendor directory from the path.
+func UnVendor(path string) string {
```
jcchavezs, line 25: Shall we cover this with tests? Also, will `/a/vendor` won't be trimmed?

Finally, would be good to explain how `/a/vendor/b` -> `/ab` is meaningful.

mrproliu: > Shall we cover this with tests?

Sure, I will create a PR for this.

> Also, will `/a/vendor` won't be trimmed?

Yes, It won't be trimmed. Because `vendor` needs to be part of the path in order to take effect.

> Finally, would be good to explain how `/a/vendor/b` -> `/ab` is meaningful.

`/b` should be meaningful, because when we do an instrument, we only care about the actual package name, such as `github.com/apache/skywalking-go/agent/core`, but the vendor management project will add a `vendor` package prefix for this, so we need to trim it.

PR #81, opened 2023-07-20
Support enhance native `mutex` as plugin

mrproliu:
Lock performance-related metrics:
image


CHANGES.md
```diff
@@ -22,6 +22,7 @@ Release Notes.
 * Support [Native SQL](https://pkg.go.dev/net/http) database client framework with [MySQL Driver](github.com/go-sql-driver/mysql).
 * Support [Logrus](https://github.com/sirupsen/logrus) log report to the backend.
 * Support [Zap](https://github.com/uber-go/zap) log report to the backend.
+* Support [Native Mutex](https://pkg.go.dev/sync) lock client framework.
```
wu-sheng, line 25:
```suggestion
* Support [Native Mutex](https://pkg.go.dev/sync) lock client framework.
```

Lock client? Is it a distributed lock solution as well?

wu-sheng: Also, you only support tracing, but not logic endpoint or metric of lock(pending, release, etc.).
Should we consider that?

mrproliu: > Lock client? Is it a distributed lock solution as well?

It's a local lock solution. Not support distribution. 

> Also, you only support tracing, but not logic endpoint or metric of lock(pending, release, etc.).
> Should we consider that?

I think the metrics should be considered, but the original lock would be triggered by Go Self and frameworks. The metrics maybe difficult to understand.

wu-sheng: I mean, how to measure a lock performance? Usually, if lock pending time is long(lock method is blocked), it mean too many competition or not enough resources pooled.

mrproliu: Yes, I think the lock time is the main metric for measuring a lock performance. We can intercept the lock begin time and finish time. 
Also, If it's using `TryLock` we can measure the failure lock count.

wu-sheng: Lock begin method should indicate the pending time before lock acquired, right?

At least 3 metrics, acquisition latency, using duration, error count(for try lock)

mrproliu: Yes.

acquisition latency: getting lock used time. 
using duration: the duration between lock and unlock time.
error count: can't getting lock count(for `TryLock`).

Am I right?

wu-sheng: Yes, would this be listed as Golang VM tab?

mrproliu: Yes, I could add to it.

mrproliu: The metrics been added to this PR.

mrproliu: The metrics been added to this PR.

wu-sheng:
Isn't lock having a name?

mrproliu:
In golang code, it's just under the "sync" package, there is no name.

wu-sheng:
Then, how to distict the different lockers? I am feeling aggregating globally is not that useful.

mrproliu:
Yes, that's why I say it maybe difficult to understand. For now, I just calculating the metrics that must in a tracing context. 
Other solution is we only calculate the lock performance when executing specific method signature(user defined), but it maybe take more CPU resource.

wu-sheng:
What do you nean in tracing context? What is IN?

mrproliu:
I mean its only could calculate the metrics when have a span(tracing context) in current goroutine.

wu-sheng:
I noticed that, but if there are more than one lock involved, it will fail, right?

wu-sheng:
This concers me, as the instrument affects the agent kernel.
Let's hold it for a little longer, to see whether we have a better solution.

tools/go-agent/config/loader.go
```diff
@@ -282,6 +282,20 @@ func (s *StringValue) GetListStringResult() []string {
 	return strings.Split(val, ",")
 }
 
+// GetListStringResultByLookup get the value from os.LookupEnv, if the env found, even it's empty
+func (s *StringValue) GetListStringResultByLookup() []string {
```
jcchavezs, line 286: Shall we test this function? I know it is very simple but I can't find a good reason for not to test it.

PR #90, opened 2023-08-10
chore: updates actions to improves CI times and resource management.

jcchavezs:
Upgrading to setup-go@v4 enables caching for deps and also allowing to build vendor folder and pass it into docker image saves dependency pull inside the docker image but also uses the dependency caching.

A next step would be to use build cache to cache test execution for go tests.


.github/workflows/publish-docker.yaml
```diff
@@ -35,9 +35,19 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: true
+      - name: Setup Golang
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.18
+      - name: Pull deps for tools/go-agent
```
jcchavezs, line 45: Small trick here, we pull deps for tools/go-agent which internally uses the dep cache and then package it into vendor module which allows to be reused in the docker build.

.github/workflows/plugin-tests.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
     timeout-minutes: 60
     steps:
       - name: Set up Go 1.18
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v4
```
jcchavezs, line 37: Takes advantage of deps caching.

jcchavezs:
I wonder how to run the kratosv2 test in local.

mrproliu:
> I wonder how to run the kratosv2 test in local.
Sorry for the late reply. 

It should same as the CI, you need to [build the plugin test tool](https://github.com/apache/skywalking-go/blob/main/.github/workflows/plugin-tests.yaml#L54) and [run the script](https://github.com/apache/skywalking-go/blob/main/.github/workflows/plugin-tests.yaml#L103).

tools/go-agent/go.mod
```diff
@@ -3,24 +3,98 @@ module github.com/apache/skywalking-go/tools/go-agent
 go 1.18
 
 require (
+	github.com/apache/skywalking-go/plugins/core v0.0.0-20230809094220-b3b15735a623
```
mrproliu, line 6: We are already using go workspace, why do we need to import package here?

wu-sheng:
Closing. No update for months. When you are ready, open a new PR.

PR #88, opened 2023-08-05
plugin grpc Unary

Alipebt:

PR #94, opened 2023-08-16
Add grpc plugin

Alipebt:
# Change
### Add grpc Streaming plugin.
* Support  ServerStream
* Support ClientStream
* Support BidirectionalStream

### Based on #88(grpc Unary plugin) 
* Support Unary


plugins/grpc/server_recvmsg_interceptor.go
```diff
@@ -0,0 +1,65 @@
...
+	return nil
+}
+
+func (h *ServerRecvMsgInterceptor) AfterInvoke(invocation operator.Invocation, result ...interface{}) error {
+	if invocation.GetContext() == nil {
+		return nil
+	}
+	span := invocation.GetContext().(tracing.Span)
+	err, ok := result[0].(error)
+	if ok && err != nil && err != io.EOF {
```
mrproliu, line 55: Why do you need to compare with the `io.EOF`, could you help to descript it to help us understand?

plugins/grpc/server_sendmsg_interceptor.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/apache/skywalking-go/plugins/core/tracing"
+)
+
+type ServerSendMsgInterceptor struct {
+}
+
+func (h *ServerSendMsgInterceptor) BeforeInvoke(invocation operator.Invocation) error {
+	ss := invocation.CallerInstance().(*nativeserverStream)
+	method := ss.s.Method()
+	if strings.HasPrefix(method, "/skywalking") {
```
mrproliu, line 33: The "/skywalking" also should be added to the "constants.go"?

test/plugins/scenarios/kratosv2/config/excepted.yml
```diff
@@ -114,7 +114,7 @@ segmentItems:
             skipAnalysis: false
             tags:
               - {key: url, value: not null}
-              - {key: RPCType, value: not null}
+              - {key: rpc.type, value: not null}
```
mrproliu, line 117: If the value of the tag is clear, please do not use `not null`.

plugins/grpc/client_closesend_interceptor.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"github.com/apache/skywalking-go/plugins/core/operator"
+	"github.com/apache/skywalking-go/plugins/core/tracing"
+)
+
+type ClientCloseSendInterceptor struct {
+}
+
+func (h *ClientCloseSendInterceptor) BeforeInvoke(invocation operator.Invocation) error {
+	endSnapshot := tracing.GetRuntimeContextValue(END_SNAPSHOT)
```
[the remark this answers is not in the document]
Alipebt: In grpc-go, client streaming starts with `NewStream`, but we use the `stream` to operate, for example:

```go
stream, err := NewStream()
go func() {
    for{
        req, err := stream. Recv()
    }
}()
go func() {
    for{
        err := stream. Send()
    }
}()
```

In order to make the span of `send` and `recv` be the child span of `NewStream`,we should get runtime context after `NewStream` span creation.

But even if `AsyncFinish()` is called in another goroutine, the `NewStream` span is still active in the context after the service ends. In order not to mix the next service with this service, we also need to record the context when `NewStream` ends. When the service ends, use it to restore the end state of NewStream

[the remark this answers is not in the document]
mrproliu: Do you mean `stream.Recv()` and `stream.Send()` is decided by the user application or grpc-go?

Also, can we use `EnhancedInstance` to enhance the streaming rather than runtime context, because it will copy all context when creating a new goroutine.

plugins/grpc/client_closesend_interceptor.go
```diff
@@ -0,0 +1,68 @@
...
+	if endSnapshot != nil {
+		tracing.ContinueContext(endSnapshot.(tracing.ContextSnapshot))
+		tracing.SetRuntimeContextValue(END_SNAPSHOT, endSnapshot)
```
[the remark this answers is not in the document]
Alipebt: `ContinueContext` is the context that needs to be continued,
`EndContext` is the context after the span ends.
But `ContinueContext` doesn't contain `EndContext`.So first we need to continue  `EndContext` in request and response,then immediately continue `ContinueContext`,In the end we still have to go back to `EndContext`. In fact, `ContinueContext` does not contain `EndContext`, so we need to set `EndContext` again after continuing `ContinueContext`, and at this time the content of `ContinueContext` has changed again, we need to replace `ContinueContext` which without `EndContext` again.

plugins/grpc/client_closesend_interceptor.go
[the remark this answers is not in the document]
Alipebt: It is decided by the user application.

Do you mean I need to use `EnhancedInstance` to set and get context?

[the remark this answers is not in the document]
mrproliu: > Do you mean I need to use EnhancedInstance to set and get context?

Yes. Also, I think it records the basic information enough(such as method name, new stream span, etc. ). And let the `NewStream` as a local span, 
`Recv` and `Send` as exit span. `Recv` and `Send` only should append to the current execution goroutine. 

Imagine the following scenario, I create streaming on a goroutine(such as initializing the system), and send/recv streaming on another goroutine. If you keep the init goroutine tracing context, it will interfere with the context of the current request. WDYT?
```
var stream protocol.Streaming
func Init(connection) {
   stream = protocol.NewStreaming(connection)
}

func HandleRequest() {
   stream.Send/Recv()
}
```

plugins/grpc/server_recvmsg_interceptor.go
Alipebt: If the peer closes sending, the return value of `stream.Recv()` will be `io.EOF`. This is the way in grpc-go to judge that the peer closes sending.

plugins/grpc/client_streaming_interceptor.go
```diff
@@ -0,0 +1,87 @@
...
+	continueSnapShot := tracing.CaptureContext()
+	span.End()
+	csEnhanced, ok := result[0].(operator.EnhancedInstance)
+	if !ok {
+		return nil
+	}
+	csEnhanced.SetSkyWalkingDynamicField(&contextData{
+		asyncSpan:        span,
+		continueSnapShot: continueSnapShot,
+		endSnapShot:      tracing.CaptureContext(),
```
mrproliu, line 83: `continueSnapshot` and `endSnapshot` are the same, please remote one.

Alipebt: > `continueSnapshot` and `endSnapshot` are the same, please remote one.

I call tracing.CaptureContext() before and after s.end() and get two snapshots. They are not the same, endsnapshot is only used after Streaming is completely over.

mrproliu: Then, please add a comment on there, I think he may make it difficult for others to understand his meaning.

mrproliu (approved):

PR #98, opened 2023-08-25
test: add a compitable version function

LinuxSuRen:


wu-sheng:
Is there any component you are going to support with this style of the version?

LinuxSuRen:
I'm trying to use sky-go in [api-testing](https://github.com/LinuxSuRen/api-testing). During the process, I found out it will add the prefix `v` no matter if the version has or not.

LinuxSuRen:
I got an error when execute the `make test` command locally. Considering this is not related to my PR. Just ignored it.

image

wu-sheng:
> I got an error when execute the `make test` command locally. Considering this is not related to my PR. Just ignored it.
> 
> image

@mrproliu PTAL. Is it worth checking whether anything is missing when tracing gRPC?

mrproliu (approved):
Generally LGTM. But It is usually decided by the version environment when we release the skywalking-go.

mrproliu:
> I got an error when execute the `make test` command locally. Considering this is not related to my PR. Just ignored it.

I think you need to more debug into it, it already running many times on the CI and works well. Please focus on the server_test.go:87, It should not throw an error, you need to check out what's happened.

mrproliu:
> Is it worth checking whether anything is missing when tracing gRPC?

For the UT, It is not related to the gRPC plugin.

LinuxSuRen:
> It is usually decided by the version environment when we release the skywalking-go.

Totally understand it. I'm wondering if it's possible and reasonable to allow user passing the version.

wu-sheng:
There are plugin tests failing, is this change breaking something?

LinuxSuRen:
> There are plugin tests failing, is this change breaking something?

Shouldn't be. This is a compitable change. Or it does not allow this change.

mrproliu:
> I'm wondering if it's possible and reasonable to allow user passing the version.

Users don't need to know about the "version" parameter. They just need to download the specified version when using it.

PR #103, opened 2023-09-22
fix: no data receive from grpc

LinuxSuRen:
## What happend
I cannot see the gRPC related data from SkyWaling native UI.

## Test
See the following screenshots with the changes:

image

I read part code lines of this project. But I'm still not familar with the picture. Please let me know if I didn't understand the related codes correctly.

By the way, I still try to figure out how to add the corresponding unit tests for these changes.

### DIFF
--- a/plugins/grpc/instrument.go
+++ b/plugins/grpc/instrument.go
@@ -69,6 +69,21 @@ func (i *Instrument) Points() []*instrument.Point {
 		},
 		{
 			PackagePath: "",
+			At: instrument.NewMethodEnhance("*ClientConnInterface", "Invoke",
+				instrument.WithArgType(0, "context.Context"),
+				instrument.WithArgType(1, "string")),
+			Interceptor: "ClientUnaryInterceptor",
+		},
+		{
+			PackagePath: "",
+			At: instrument.NewMethodEnhance("*ClientConnInterface", "NewStream",
+				instrument.WithArgType(0, "context.Context"),
+				instrument.WithResultCount(2),
+				instrument.WithResultType(0, "ClientStream")),
+			Interceptor: "ClientStreamingInterceptor",
+		},
+		{
+			PackagePath: "",
 			At: instrument.NewMethodEnhance("*Server", "handleStream",
 				instrument.WithArgsCount(3),
 				instrument.WithArgType(0, "transport.ServerTransport")),
--- a/tools/go-agent/instrument/logger/frameworks/buildin.go
+++ b/tools/go-agent/instrument/logger/frameworks/buildin.go
@@ -1,0 +1,42 @@
+package frameworks
+
+import (
+	"github.com/apache/skywalking-go/tools/go-agent/instrument/plugins/rewrite"
+	"github.com/dave/dst"
+	"github.com/dave/dst/dstutil"
+)
+
+type Buildin struct {
+}
+
+func NewBuildin() *Buildin {
+	return &Buildin{}
+}
+
+func (l *Buildin) Name() string {
+	return "buildin"
+}
+
+func (l *Buildin) PackagePaths() map[string]*PackageConfiguration {
+	return map[string]*PackageConfiguration{"log": {NeedsHelpers: true, NeedsVariables: true, NeedsChangeLoggerFunc: true}}
+}
+
+func (l *Buildin) AutomaticBindFunctions(fun *dst.FuncDecl) string {
+	return ""
+}
+
+func (l *Buildin) GenerateExtraFiles(pkgPath, debugDir string) (info []*rewrite.FileInfo, err error) {
+	return
+}
+
+func (l *Buildin) CustomizedEnhance(path string, curFile *dst.File, cursor *dstutil.Cursor, allFiles []*dst.File) (cache map[string]string, ok bool) {
+	return
+}
+
+func (l *Buildin) InitFunctions() []*dst.FuncDecl {
+	return nil
+}
+
+func (l *Buildin) InitImports() []*dst.ImportSpec {
+	return nil
+}
--- a/tools/go-agent/instrument/logger/instrument.go
+++ b/tools/go-agent/instrument/logger/instrument.go
@@ -44,6 +44,7 @@ import (
 var logFrameworks = []frameworks.LogFramework{
 	frameworks.NewLogrus(),
 	frameworks.NewZap(),
+	// frameworks.NewBuildin(),
 }
 
 //go:embed context.go


wu-sheng:
I can see gRPC plugin exists in dev only with tests. Have you cross verified with that?

wu-sheng:
FYI @Alipebt as you added gRPC plugin.

LinuxSuRen:
> I can see gRPC plugin exists in dev only with tests. Have you cross verified with that?

Could be more specific? I just checked the result before and after with this PR.

By the way, I checked the interface name of gRPC client. It is `type ClientConnInterface interface`. I'm not sure how `*ClientConn"` works. Maybe I missed something such as version.

https://github.com/grpc/grpc-go/blob/fe0dc2275d7dfdd61c6915bd3a4ec5bff37963e3/clientconn.go#L618

Any way, please let me know if anyone konw the difference.

wu-sheng:
I mean this is the test, https://github.com/apache/skywalking-go/blob/b3e0cf3c66e3a4d9010d050b6a2015c4014b8a33/.github/workflows/plugin-tests.yaml#L88

All contributed plugins, including the ones made by the maintainers, require tests with various declared versions, as well as codes to help further users/contributora know what are tested.

This doc could help you to understand the tests. https://skywalking.apache.org/docs/skywalking-go/next/en/development-and-contribution/write-plugin-testing/

All these make maintainers don't debug the plugin for users and contributors. You could cross check, do local verification, and submit a patch or an enhancement with new test codes.

tools/go-agent/instrument/logger/frameworks/buildin.go
```diff
@@ -0,0 +1,42 @@
+package frameworks
+
+import (
+	"github.com/apache/skywalking-go/tools/go-agent/instrument/plugins/rewrite"
+	"github.com/dave/dst"
+	"github.com/dave/dst/dstutil"
+)
+
+type Buildin struct {
```
mrproliu, line 9: Why you need to add this? It's not related to the grpc plugin.

LinuxSuRen: Sorry about this. I forgot the remove the unrelated lines.

plugins/grpc/instrument.go
```diff
@@ -67,6 +67,21 @@ func (i *Instrument) Points() []*instrument.Point {
 				instrument.WithResultType(0, "ClientStream")),
 			Interceptor: "ClientStreamingInterceptor",
 		},
+		{
+			PackagePath: "",
+			At: instrument.NewMethodEnhance("*ClientConnInterface", "Invoke",
```
Alipebt, line 72: The interface cannot be intercepted, and I think intercepting `ClientConnInterface` is not effective.

`ClientConn` implements this interface, and when user make RPC calls, the Invoke and `NewStream` methods implemented by `ClientConn` are executed.

Alipebt:
What data are you referring to? I think the picture you provided is gRPC related data.

wu-sheng:
> What data are you referring to? I think the picture you provided is gRPC related data.

I am wondering that too. @LinuxSuRen Could you explain what is your point of asking? The agent would not intercept your all codes, or not have to be the codes(APIs) you are calling. The meaning of exit spans(your screenshots) measures the performance of remote peer(server side and the network) as much as possible. 
Also, for the distributed tracing, the trace would continue if the remote server has been instrumented. You could check the gRPC calling you made, there is a `sw8` header injected by the agent to make the trace continuous.

wu-sheng:
This is a background you should learn about distributed tracing, https://research.google/pubs/pub36356/.
And we are not measuring the performance of lines of codes unless you begin to touch the part called profiling.

plugins/grpc/instrument.go
LinuxSuRen: Let me do more tests. I didn't get the tracing data of gRPC before. Below is my usage.

https://github.com/LinuxSuRen/api-testing/blob/58905fdd8f42b9f988f6312c7fe72034e0e12fc7/pkg/runner/grpc.go#L329

LinuxSuRen:
> What data are you referring to? I think the picture you provided is gRPC related data.

I can see the data from that picture when I build `skywalking-go-agent` with this PR.

hi @wu-sheng , for the background of my project. The workflow is below:

A -> HTTP API -> call the extension which is a gRPC server.

So, sk could get the gRPC tracing data if I understand correctly. Currently, I only inject sk into the gRPC client side.

wu-sheng:
Follow this if that is your point, https://github.com/apache/skywalking-go/pull/103#discussion_r1334349474

plugins/grpc/instrument.go
LinuxSuRen: Sure. I need more time to make sure my use case is correct. I will add the corresponding tests after that.

LinuxSuRen: I feel very confused. The e2e test seem correct, but it cannot work in my case. So, I tend to close this PR until I can figure it out.

By the way, I didn't find a way to control if sending the data to SK server or not. For some use cases (at least in my cas), SkyWalking is not required for the installation. So, I'm wonderfing if there is a way to control it.

I found a possible solution is that. Read an environment variable `SW_AGENT_ENABLED`, return `false` if its value is `false`.

https://github.com/apache/skywalking-go/blob/b3e0cf3c66e3a4d9010d050b6a2015c4014b8a33/tools/go-agent/instrument/reporter/instrument.go#L47

wu-sheng: > By the way, I didn't find a way to control if sending the data to SK server or not

No, you can't. That is kernel responsibility, including connection management, TLS, traffic limitation, etc.

> For some use cases (at least in my cas), SkyWalking is not required for the installation

What do you mean some cases? Do you mean some URL or protocol? Or something else? You want to compile the agent into your binary, but not using it?

LinuxSuRen: > What do you mean some cases? Do you mean some URL or protocol? Or something else? You want to compile the agent into your binary, but not using it?

Yes. Users might only want to install the tool instead of other optional component. For example my test tool, the following could start a server:

`atest server`

It will output the following errors if no SK server given. I'm thinking about if it's possible to avoid printing the errors.

`skywalking-go 2023/09/25 14:58:14 open stream error rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 0.0.0.0:11800: connect: connection refused"`

wu-sheng: In the previous discussion, we were not considering a general use tool would adopt this agent. So, the old proposal is going to package two binaries, one with the agent, the other is not. This should be good enough for a business service.

@mrproliu Do we have the mechanism for the runtime configuration for now? If so, maybe we could consider add one to disable the agent.

LinuxSuRen: Thanks for your explanation. Consider my project is very small and simple, having mutiple distribution is not a good choice. As you mentioned, it's not problem for those business vendors.

wu-sheng: Let's focus on the feature at here.

LinuxSuRen:
Close this PR until I could figure it out.

LinuxSuRen:
Finally, I figured out the reason. It's too embarrassing that I used `0.0.2` which does have not the `gRPC` plugin.

PR #122, opened 2023-10-12
feat: update Makefile to support build arm64 platform

YenchangChan:
I wan't to build skywalking-go agent on arm64 platform, but I found that Makefile defines ARCH=amd64 with dead code, so I update the Makefile, that I can compile arm64 platform with:
```
make build ARCH=arm64
```


mrproliu (approved):
Thanks

PR #134, opened 2023-10-24
feat: support setting a discard type of reporter

LinuxSuRen:
As mentioned in the title, users could discard the report data via environment variable setting.

### DIFF
--- a/tools/go-agent/Makefile
+++ b/tools/go-agent/Makefile
@@ -51,4 +51,4 @@ build: linux darwin windows
 .PHONY: $(PLATFORMS)
 $(PLATFORMS):
 	mkdir -p $(OUT_DIR)
-	GOOS=$(os) GOARCH=$(ARCH) $(GO_BUILD) $(GO_BUILD_FLAGS) -ldflags "$(GO_BUILD_LDFLAGS)" -o $(OUT_DIR)/$(BINARY)-$(VERSION)-$(os)-$(ARCH) ./cmd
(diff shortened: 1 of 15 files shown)


tools/go-agent/Makefile
```diff
@@ -51,4 +51,4 @@ build: linux darwin windows
 .PHONY: $(PLATFORMS)
 $(PLATFORMS):
 	mkdir -p $(OUT_DIR)
-	GOOS=$(os) GOARCH=$(ARCH) $(GO_BUILD) $(GO_BUILD_FLAGS) -ldflags "$(GO_BUILD_LDFLAGS)" -o $(OUT_DIR)/$(BINARY)-$(VERSION)-$(os)-$(ARCH) ./cmd
+	CGO_ENABLED=0 GOOS=$(os) GOARCH=$(ARCH) $(GO_BUILD) $(GO_BUILD_FLAGS) -ldflags "$(GO_BUILD_LDFLAGS)" -o $(OUT_DIR)/$(BINARY)-$(VERSION)-$(os)-$(ARCH) ./cmd
```
LinuxSuRen, line 54: Will occur the `GLIC` relevant issues.

mrproliu:
Can you provide more scenarios to illustrate when users need to use it? If they don't need to collect data and can perform operations during compilation, it will result in better performance.

LinuxSuRen:
There are two important reason.

* Considering SkyWalking is not a MUST component all the time. I don't expect that it prints error logs when no SkyWalking installed.
* Maintaining the double number of binary files is an annoying thing. Let's assume one project has different binary files against Windows, Linux, macOS. If there is no such switch to turn off reporting, then users need to publish 6 packages everytime instead of 3. Even wrose if arm64 is required, users might need to mtaintain 12 packages.

mrproliu (approved):

mrproliu (commented):
Then, could you please add this feature in the [CHANGE.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md)

mrproliu (approved):
Thanks!

PR #136, opened 2023-10-25
fix: users can not use async api in toolkit-trace

Alipebt:
## Description

In the process of writing my blog, I found that Async API in manual APIs are not available to users because users cannot pass Span into other goroutines. I accidentally overlooked this before.

Then I referred to JAVA and found that it provides users with a `spanRef` to use async api. (I've defined this `spanRef`, but it's currently no use)
Screenshot_20231025_122044

So I changed the async api.

## Change
- fix Async API in manual APIs
- Supplement cross goroutine async test

### DIFF
--- a/plugins/trace-activation/create_entryspan_intercepter.go
+++ b/plugins/trace-activation/create_entryspan_intercepter.go
@@ -27,13 +27,20 @@ type CreateEntrySpanInterceptor struct {
 }
 
 func (h *CreateEntrySpanInterceptor) BeforeInvoke(invocation operator.Invocation) error {
-	operationName := invocation.Args()[0].(string)
-	var extractor func(headerKey string) (string, error) = invocation.Args()[1].(trace.ExtractorRef)
-	s, err := tracing.CreateEntrySpan(operationName, extractor)
-	invocation.DefineReturnValues(s, err)
 	return nil
 }
 
 func (h *CreateEntrySpanInterceptor) AfterInvoke(invocation operator.Invocation, result ...interface{}) error {
+	operationName := invocation.Args()[0].(string)
+	var extractor func(headerKey string) (string, error) = invocation.Args()[1].(trace.ExtractorRef)
+	s, err := tracing.CreateEntrySpan(operationName, extractor)
+	if err != nil {
+		return nil
+	}
+	enhancced, ok := result[0].(operator.EnhancedInstance)
+	if !ok {
+		return nil
+	}
+	enhancced.SetSkyWalkingDynamicField(s)
 	return nil
 }
--- a/plugins/trace-activation/create_entryspan_intercepter.go
+++ b/plugins/trace-activation/create_entryspan_intercepter.go
@@ -36,6 +36,7 @@ func (h *CreateEntrySpanInterceptor) AfterInvoke(invocation operator.Invocation,
 	s, err := tracing.CreateEntrySpan(operationName, extractor)
 	if err != nil {
 		invocation.DefineReturnValues(nil, err)
+		return nil
 	}
 	enhancced, ok := result[0].(operator.EnhancedInstance)
 	if !ok {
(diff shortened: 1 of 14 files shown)


plugins/trace-activation/create_localspan_intercepter.go
```diff
@@ -26,12 +26,16 @@ type CreateLocalSpanInterceptor struct {
...
 func (h *CreateLocalSpanInterceptor) BeforeInvoke(invocation operator.Invocation) error {
-	operationName := invocation.Args()[0].(string)
-	s, err := tracing.CreateLocalSpan(operationName)
-	invocation.DefineReturnValues(s, err)
 	return nil
 }
 
 func (h *CreateLocalSpanInterceptor) AfterInvoke(invocation operator.Invocation, result ...interface{}) error {
+	operationName := invocation.Args()[0].(string)
+	s, _ := tracing.CreateLocalSpan(operationName)
```
mrproliu, line 34: Same here, the error message cannot be ignored.

plugins/trace-activation/create_exitspan_intercepter.go
```diff
@@ -27,14 +27,21 @@ type CreateExitSpanInterceptor struct {
...
-	s, err := tracing.CreateExitSpan(operationName, peer, injector)
-	invocation.DefineReturnValues(s, err)
 	return nil
 }
 
 func (h *CreateExitSpanInterceptor) AfterInvoke(invocation operator.Invocation, result ...interface{}) error {
+	operationName := invocation.Args()[0].(string)
+	peer := invocation.Args()[1].(string)
+	var injector func(headerKey, headerValue string) error = invocation.Args()[2].(trace.InjectorRef)
+	s, err := tracing.CreateExitSpan(operationName, peer, injector)
```
mrproliu, line 37: Same here, the error message cannot be ignored.

plugins/trace-activation/create_entryspan_intercepter.go
```diff
@@ -27,13 +27,20 @@ type CreateEntrySpanInterceptor struct {
...
-	var extractor func(headerKey string) (string, error) = invocation.Args()[1].(trace.ExtractorRef)
-	s, err := tracing.CreateEntrySpan(operationName, extractor)
-	invocation.DefineReturnValues(s, err)
 	return nil
 }
 
 func (h *CreateEntrySpanInterceptor) AfterInvoke(invocation operator.Invocation, result ...interface{}) error {
+	operationName := invocation.Args()[0].(string)
+	var extractor func(headerKey string) (string, error) = invocation.Args()[1].(trace.ExtractorRef)
+	s, err := tracing.CreateEntrySpan(operationName, extractor)
```
mrproliu, line 36: The error message cannot be ignored, we should re-write the result when the error is not nil.

toolkit/trace/span.go
```diff
@@ -21,5 +21,5 @@ type ExtractorRef func(headerKey string) (string, error)
 
 type InjectorRef func(headerKey, headerValue string) error
```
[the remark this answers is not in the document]
Alipebt: Ok，and how to define the file name appropriately？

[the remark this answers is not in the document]
mrproliu: using the `api.go` is enough.

plugins/trace-activation/create_entryspan_intercepter.go
```diff
@@ -27,13 +27,20 @@ type CreateEntrySpanInterceptor struct {
...
+	if err != nil {
+		invocation.DefineReturnValues(nil, err)
```
mrproliu, line 38: Forget `return`?

plugins/trace-activation/create_exitspan_intercepter.go
```diff
@@ -27,14 +27,21 @@ type CreateExitSpanInterceptor struct {
...
+	if err != nil {
+		invocation.DefineReturnValues(nil, err)
```
mrproliu, line 39: Forget `return`?

mrproliu (dismissed):

mrproliu (approved):

PR #104, opened 2023-09-23
Provide APIs to user and support correlationContext

Alipebt:
# Add
- Provide APIs to users
- support get/set correlation context value
- Add documents related these APIs.
- Add tests for toolkit
- Add the `toolkit` option in config.
Due to the interception of packages by skywalking-go, it defaults to fetching them remotely (using `go mod tidy`). However, during testing, `toolkit` only exists locally, so `go mod tidy` will be unable to fetch the remote package for `toolkit`. Therefore, I added a `toolkit` field in the config file. If this field is determined to be true, the package path will be changed to the local one using `replace`, avoiding the download and error reported by `go mod tidy`.

realted to apache/skywalking#11333

PR #150, opened 2023-11-15
Fix wrong plugin structure

lujiajing1126:
In https://github.com/apache/skywalking-go/pull/104, trace related APIs are introduced for developers.

But the plugin strucuture is not correct, this leads to empty version in the `VersionChecker`

```go
func (i *Instrument) VersionChecker(version string) bool {
        // the version is always empty
	return true
}
```

Besides, in order to make E2E work in PR #104, the result of version parser is ignored. This make toolkit APIs work by change. So even if we use an official release, e.g. `github.com/apache/skywalking-go v0.3.0`. The version is empty.

So in this PR, 

1. Revert error handling after version parse, and tolerate specific error 
2. Change plugin dir for trace-activation plugin


tools/go-agent/instrument/plugins/instrument.go
```diff
@@ -107,9 +108,8 @@ func (i *Instrument) CouldHandle(opts *api.CompileOptions) bool {
 		// check the version of the framework could handler
 		version, err := i.tryToFindThePluginVersion(opts, ins)
 		if err != nil {
-			// Local package (e.g. replaced toolkit) does not have version.
-			// So when the version is not detected, we should not skip it.
-			logrus.Warnf("the plugin %s %s", ins.Name(), err)
+			logrus.Warnf("ignore the plugin %s, because %s", ins.Name(), err)
+			continue
```
mrproliu, line 112: I remember this is used to handle the check for missing version information when `replacing` in `go.mod`, @Alipebt could you help to re-check?

lujiajing1126: > I remember this is used to handle the check for missing version information when `replacing` in `go.mod`, @Alipebt could you help to re-check?

I found the solution is not correct. Let me find another way

lujiajing1126: If we skip all args in the loop, shall we return an error?

mrproliu: I think a warm log is enough, an error will let the compile phase failure.

lujiajing1126: > I think a warm log is enough, an error will let the compile phase failure.

image

I meant here. The only possbility to reach here is no go sources.

mrproliu: No, we don't need to add the error here. If there is no go file, then it will not have any enhancements. If an error is returned here, it will cause useless error log output in the upper layer.

mrproliu (approved):

PR #153, opened 2023-11-29
fix: limit the bytes of redis args

rfyiamcool:
### summary

limit the bytes of redis args.

### DIFF
--- a/docs/en/agent/plugin-configurations.md
+++ b/docs/en/agent/plugin-configurations.md
@@ -4,6 +4,6 @@
 |--------------------------------|-------------------------------------------------------|---------------|----------------------------------------------------------------|
 | http.server_collect_parameters | SW_AGENT_PLUGIN_CONFIG_HTTP_SERVER_COLLECT_PARAMETERS | false         | Collect the parameters of the HTTP request on the server side. |
 | mongo.collect_statement        | SW_AGENT_PLUGIN_CONFIG_MONGO_COLLECT_STATEMENT        | false         | Collect the statement of the MongoDB request.                  |
-| redis.max_args_bytes | SW_AGENT_PLUGIN_CONFIG_REDIS_MAX_ARGS_BYTES | 1024 | Limit the bytes size of redis args request.                  |
 | sql.collect_parameter          | SW_AGENT_PLUGIN_CONFIG_SQL_COLLECT_PARAMETER          | false         | Collect the parameter of the SQL request.                      |
+| redis.max_args_bytes           | SW_AGENT_PLUGIN_CONFIG_REDIS_MAX_ARGS_BYTES           | 1024          | Limit the bytes size of redis args request.                    |
 | reporter.discard               | SW_AGENT_REPORTER_DISCARD                             | false         | Discard the reporter.                                          |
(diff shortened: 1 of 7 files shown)


plugins/go-redisv9/hook.go
```diff
@@ -187,3 +187,13 @@ func getKey(args []interface{}) string {
 	}
 	return key
 }
+
+const maxArgsLength = 1024
```
wu-sheng, line 191: Hard code the limitation? I don't think this is a good practice.

Please follow this pattern, https://github.com/apache/skywalking-go/blob/main/tools/go-agent/config/agent.default.yaml

rfyiamcool: done, please cr again 😁

docs/en/agent/plugin-configurations.md
```diff
@@ -4,5 +4,6 @@
 |--------------------------------|-------------------------------------------------------|---------------|----------------------------------------------------------------|
 | http.server_collect_parameters | SW_AGENT_PLUGIN_CONFIG_HTTP_SERVER_COLLECT_PARAMETERS | false         | Collect the parameters of the HTTP request on the server side. |
 | mongo.collect_statement        | SW_AGENT_PLUGIN_CONFIG_MONGO_COLLECT_STATEMENT        | false         | Collect the statement of the MongoDB request.                  |
+| redis.max_args_bytes | SW_AGENT_PLUGIN_CONFIG_REDIS_MAX_ARGS_BYTES | 1024 | Limit the bytes size of redis args request.                  |
```
wu-sheng, line 7: Please align the order here and config file.
This should be after sql config.

rfyiamcool:
> Could you add this feature into the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md)?

done. 😁

mrproliu:
Please fix the CI.

rfyiamcool:
> Please fix the CI.

I don't know how to fix this problem. 😅  give me some Pointers.

wu-sheng:
CI had logs, you should be able to read them, and CI is able to run locally.

wu-sheng:
Did you test the codes you changed? It seems it doesn't work and runnable actually.

https://github.com/apache/skywalking-go/actions/runs/7044082174/job/19171487594?pr=153

tools/go-agent/instrument/plugins/enhance_config.go
```diff
@@ -215,7 +215,7 @@ func (f *ConfigField) GenerateAssignFieldValue(varName string, field, path []str
 		getFromEnvStr = fmt.Sprintf("if v := tools.GetEnvValue(%q); v != \"\" { result = v };", pluginConfig.EnvKey)
 	}
 	parseResStr := ""
-	parseErrorMessage := "cannot parse the config " + fieldKeyPathStr + ": err.Error()"
+	parseErrorMessage := fmt.Sprintf(`"cannot parse the config %s: " + err.Error()`, fieldKeyPathStr)
```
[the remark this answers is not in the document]
rfyiamcool: When run unit test, always throw `syntax errors`.  this is a problem with the generated panic syntax, the string does not have quotes.

image


```go
panic(cannot parse the config redis: err.Error())
```

Modified effect.

```go
panic("cannot parse the config redis: " + err.Error())
```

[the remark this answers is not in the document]
rfyiamcool: Is that right? 😁

before

```
	parseErrorMessage := fmt.Sprintf(`"cannot parse the config %s: " + err.Error()`, fieldKeyPathStr)
```

after.

```
	parseErrorMessage := fmt.Sprintf(`"cannot parse the config %s: " + %s`, fieldKeyPathStr, "err.Error()")
```

mrproliu (approved):
LGTM.

PR #154, opened 2023-11-30
feat: add client ip in gin

rfyiamcool:
### summary

1. add remoteip and clientip to span.tags in gin.
2. ut.

### DIFF
--- a/plugins/core/tracing/span.go
+++ b/plugins/core/tracing/span.go
@@ -50,6 +50,8 @@ type Tag string
 
 const (
 	TagURL             = "url"
+	TagRemoteAddr      = "remote_addr"
+	TagClientIP        = "client_ip"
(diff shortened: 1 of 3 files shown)


wu-sheng:
IPs are not typically tavs for SkyWalking. What do you want witb this?
Remote IP is peer already. Client IP belongs to instance properties. You should not report these repeatedly.

rfyiamcool:
> IPs are not typically tavs for SkyWalking. What do you want witb this? Remote IP is peer already. Client IP belongs to instance properties. You should not report these repeatedly.

skywalking records the ip address of each instance, but he does not record the 'x-forwarded-for' address ?

wu-sheng:
Tracing is not to collect all, but only necessary.
So, if you want to add more, feel free to use toolkit.

PR #156, opened 2023-12-18
Changing intercept point for gin, make sure interfaces could be group…

dylanforest:
Changing intercept point for gin, make sure interfaces could be grouped when params defined in relativePath.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Release Notes.
 #### Plugins
 * Support setting a discard type of reporter.
 * Add `redis.max_args_bytes` parameter for redis plugin.
+* Changing intercept point for gin, make sure interfaces could be grouped when params defined in relativePath.
 
 #### Documentation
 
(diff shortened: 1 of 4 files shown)


dylanforest:
Is this CI failure  a problem? I think it's not code's fault.

mrproliu (approved):

wu-sheng:
> Is this CI failure a problem? I think it's not code's fault.

I rerun the CIs.

dylanforest:
> > Is this CI failure a problem? I think it's not code's fault.
> 
> I rerun the CIs.
Looks like it fails again

mrproliu:
It looks like GitHub Action caused the failure, we can only run a few more times to try. :(

dylanforest:
Sad...

wu-sheng:
@mrproliu Is there some env changed?

dylanforest:
Same problems everytime, maybe some settings wrong？

dylanforest:
Any news?

wu-sheng:
> Any news?

We had fixed this, and now this is able to merge.

PR #157, opened 2023-12-19
Fix macOS CI Testing

mrproliu:


wu-sheng (dismissed):

wu-sheng:
Plugin tests seem not working.

mrproliu:
It usually needs to wait for a while.

wu-sheng (approved):

PR #162, opened 2024-01-21
Fix enhance param error when there are multiple params

CodePrometheus:
### **summary**
When I intercepted the following method, I found that the parameter passed to skywalking is an array, and only the first parameter can be obtained.
For example, can only get ctx, A, C and msg.
```go
func (ch *Ch) Test(ctx context.Context, A, B string, C, D bool, msg Publishing) (*Dc, error) {}
```
```go
// tools/go-agent/tools/enhancement.go:63
func EnhanceParameterNames(fields *dst.FieldList, fieldType FieldListType) []*ParameterInfo {
	if fields == nil {
		return nil
	}
	result := make([]*ParameterInfo, 0)
	for i, f := range fields.List {
		var defineName string
		switch fieldType {
		case FieldListTypeParam:
			defineName = fmt.Sprintf("skywalking_param_%d", i)
		case FieldListTypeResult:
			defineName = fmt.Sprintf("skywalking_result_%d", i)
		case FieldListTypeRecv:
			defineName = fmt.Sprintf("skywalking_recv_%d", i)
		}
		if len(f.Names) == 0 {
			f.Names = []*dst.Ident{{Name: defineName}}
			result = append(result, newParameterInfo(defineName, f.Type))
		} else {
			for _, n := range f.Names {
				if n.Name == "_" {
					*n = *dst.NewIdent(defineName)
					break
				}
			}

			for _, n := range f.Names {
				result = append(result, newParameterInfo(n.Name, f.Type))
                                 // If there are multiple parameters in f.Names, only the first element of the array will be processed here
				break
			}
		}
	}
	return result
}
```


go.work.sum
```diff
@@ -1759,6 +1760,7 @@ golang.org/x/text v0.11.0 h1:LAntKIrcmeSKERyiOh0XMV39LXS8IE9UL2yP7+f5ij4=
 golang.org/x/time v0.0.0-20200416051211-89c76fbcd5d1/go.mod h1:tRJNPiyCQ0inRvYxbN9jk5I+vvW/OXSQhTDSoE431IQ=
 golang.org/x/time v0.0.0-20210611083556-38a9dc6acbc6/go.mod h1:tRJNPiyCQ0inRvYxbN9jk5I+vvW/OXSQhTDSoE431IQ=
 golang.org/x/time v0.1.0/go.mod h1:tRJNPiyCQ0inRvYxbN9jk5I+vvW/OXSQhTDSoE431IQ=
+golang.org/x/time v0.5.0 h1:o7cqy6amK/52YcAKIPlM3a+Fpj35zvRj2TP+e1xFSfk=
```
mrproliu, line 1763: Why you need to modify the go.mod? It seems no dependency update.

tools/go-agent/tools/enhancement.go
```diff
@@ -88,7 +88,6 @@ func EnhanceParameterNames(fields *dst.FieldList, fieldType FieldListType) []*Pa
 
 			for _, n := range f.Names {
 				result = append(result, newParameterInfo(n.Name, f.Type))
-				break
```
mrproliu, line 91: Could you help to add more test case on this?

mrproliu (approved):

mrproliu (changes requested):
Please update this PR by comments, thanks.

go.work.sum
CodePrometheus: Maybe it was caused by my executing `go mod tidy` locally.

mrproliu: Then, please restore to the upstream version.

tools/go-agent/tools/enhancement.go
CodePrometheus: done~

mrproliu (approved):
Once the CI success, it's LGTM. Thanks.

CodePrometheus:
Thank you for your time as well, Teacher Liu.

PR #163, opened 2024-01-23
Add support ignore suffix

CodePrometheus:
### summary
add support ignore suffix

### what
Provide users with the ability to ignore the span for the operationName dimension.

### why
In the actual usage environment, there are a batch of requests such as /xx.jpg. We can reduce the signal-to-noise ratio of the collection to better analyze the trace.

### how
I added `agent.ignore_suffix` in the default configuration file, which will add a layer of judgment when creating Entry, Local and Exit spans.

### additional
There is no test code added for the time being. Since I am considering the operationName dimension, I would like to discuss the topic [11807](https://github.com/apache/skywalking/discussions/11807) with community members~


mrproliu (approved):
LGTM

PR #165, opened 2024-01-28
feature: add support for AMQP

CodePrometheus:
Good evening friends!

[Amqp](https://github.com/rabbitmq/amqp091-go) is the Go Amqp Client Library officially maintained by the rabbitmq team.

I would be very grateful if a community member could review my code and provide some feedback.

PR #169, opened 2024-02-13
feature: add support for rocketMQ

CodePrometheus:
### summary

[RockeMQ](https://github.com/apache/rocketmq-client-go) is a native go client maintained by the apache rocketMQ team.

### test report
I intercepted the `SendSync`, `SendAsync`, and `SendOneWay` methods in the producer, as shown in the figure below.

case1
case2
case3


wu-sheng:
Could you confirm, is the further process driven by the message going to be in the same context?

CodePrometheus:
Ref https://github.com/apache/skywalking/pull/11877

CodePrometheus:
> Could you confirm, is the further process driven by the message going to be in the same context?

I think so, so I added headKey in properties, I'm not sure if I'm articulating the point you want to ask about
```
type Message struct {
  properties map[string]string
}
```

wu-sheng:
I mean, the consumer codes, it would do business operations, such as accessing SQL database, cache, which are supported to be traced. Are they im the same trace? 

The reason I am asking for is, message consuming sometimes has a kind of async mechanisms, which may change the tracing context incorrectly. It is better to recheck with demo codes.

CodePrometheus:
I double checked, they are not in the same trace. The chain between producer and consumer is independent, as shown in the diagram, when I make additional http requests within the consumer code. I think these other business operations in consumer will not affect the tracing context, maybe I have overlooked something, please kindly give me your advice

image

wu-sheng:
The use case should be
1. A produces a message
2. B consumes the message
3. B uses this message to continue the processes.

These three should be in one trace. If 1 and 2 are, 3 is not, it means you miss one local span to wrap the 2 and 3 in a context. Java agent plugin should have covered this case.

CodePrometheus:
image
image


The further process driven by the message will be in the same context, the context can be propagated and intercepted as it is passed down the execution chain

wu-sheng:
Good to me. Let's wait for @mrproliu review.

test/plugins/scenarios/rocketmq/main.go
```diff
@@ -0,0 +1,246 @@
...
+func main() {
+	route := http.NewServeMux()
+	route.HandleFunc("/execute", func(res http.ResponseWriter, req *http.Request) {
+		_, _ = res.Write([]byte("execute success"))
+	})
+	route.HandleFunc("/health", func(writer http.ResponseWriter, request *http.Request) {
+		_, _ = writer.Write([]byte("ok"))
+	})
+
+	TestProCon()
```
mrproliu, line 55: Please invoke this method in the execute handler, it more like the real scene.

test/plugins/scenarios/rocketmq/config/excepted.yml
```diff
@@ -0,0 +1,181 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+segmentItems:
+  - serviceName: rocketmq
+    segmentSize: ge 9
+    segments:
+      - segmentId: not null
```
mrproliu, line 21: Why all segments only have single span? I think some span should belong to the same segment.

mrproliu (commented):
Also, please fix the CI.

plugins/rocketmq/producer/send_one_way.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/apache/skywalking-go/plugins/core/operator"
+	"github.com/apache/skywalking-go/plugins/core/tracing"
+)
+
+type SendOneWayInterceptor struct {
+}
+
+func (s *SendOneWayInterceptor) BeforeInvoke(invocation operator.Invocation) error {
+	return SimpleProducerInterceptor(invocation)
```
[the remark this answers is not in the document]
CodePrometheus: What I want to express is that sendSync and SendOneWay share a BeforeInvoke, `GeneralProducerBeforeInvoke` What do you think

test/plugins/scenarios/rocketmq/config/excepted.yml
```diff
@@ -0,0 +1,181 @@
...
+        spans:
+          - operationName: RocketMQ/sw-topic/Producer/Callback
+            parentSpanId: -1
+            spanId: 0
+            spanLayer: MQ
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 38
+            isError: false
+            spanType: Exit
```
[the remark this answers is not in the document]
CodePrometheus: There are two callback spans, the first is [exit](https://github.com/apache/skywalking-go/pull/169/files#diff-8f790a9cc27657931636c2242e8089661e539d72262132c1837cce958b2fd40fR75) and the second is [local](https://github.com/apache/skywalking-go/pull/169/files#diff-8f790a9cc27657931636c2242e8089661e539d72262132c1837cce958b2fd40fR90)

test/plugins/scenarios/rocketmq/main.go
CodePrometheus: Because I encountered the same problem before during development https://github.com/apache/skywalking-go/pull/165#issuecomment-1935516907

plugins/rocketmq/producer/send_one_way.go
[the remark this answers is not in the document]
mrproliu: Sure, this name is good.

test/plugins/scenarios/rocketmq/config/excepted.yml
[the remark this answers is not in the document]
mrproliu: Could you help to do more explain the callback spans. When should be exit or local?

test/plugins/scenarios/rocketmq/main.go
mrproliu: I mean, you should adding send message(producer) to the http handler, the consumer should other thing. Because you have two entry span, so the span name been rewrited.

test/plugins/scenarios/rocketmq/config/excepted.yml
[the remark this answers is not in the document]
CodePrometheus: [async](https://github.com/apache/skywalking-go/pull/169/files#diff-cbae1ae787d2022d1035230dccde1263fff109b1d98c4bab947b0e9259bd4da1R42) My understanding of rocketMQ's Async is that after the producer sends the data, I create an Exit and do not need to wait for the return, instead, receive the server response through the callback interface, at this time, I create a Local span to represent the result of the callback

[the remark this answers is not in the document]
mrproliu: Ok, then please update the excepted files by comment, to make its more clear.

test/plugins/scenarios/rocketmq/main.go
CodePrometheus: very sorry, I have been stuck on this question for a long time, based on Teacher Wu and your suggestions, I have tried it like this, the UI is displayed as follows

image

[excepted.yml](test/plugins/scenarios/rocketmq/config/excepted.yml), If there are still problems, please prompt me with more info, thanks

plugins/rocketmq/consumer/consumer.go
```diff
@@ -0,0 +1,92 @@
...
+			tracing.WithLayer(tracing.SpanLayerMQ),
+			tracing.WithComponent(rmqConsumerComponentID),
+			tracing.WithTag(tagMQMsgID, msg.MsgId),
+			tracing.WithTag(tagMQOffsetMsgID, msg.OffsetMsgId),
+		)
+		if err != nil {
+			return err
+		}
+	}
+	span.Tag(tracing.TagMQTopic, topic)
```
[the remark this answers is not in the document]
CodePrometheus: done

CodePrometheus:
Sorry for the recurring problem on CI/Tests. My `golangci-lint` attempt on the local mac seems to be inconsistent with that on github. Is there any additional configuration required besides `make lint`?

wu-sheng:
@mrproliu Any idea?

mrproliu:
> Sorry for the recurring problem on `CI / Tests (pull_request) `. My `golangci-lint` attempt on the local mac seems to be inconsistent with that on github. Is there any additional configuration required besides `make lint`?

For now, you can delete the local lint command, then, it could be download the specific lint command by `make lint` automatically.

test/plugins/scenarios/rocketmq/main.go
mrproliu: image
They have same span name, I think I could make user confuse, should we change names?

CodePrometheus: ok, I will use different operationName to identify async and its corresponding callback.

CodePrometheus:
> For now, you can delete the local lint command, then, it could be download the specific lint command by `make lint` automatically.

Thanks, it works, from yesterday's [test](https://github.com/apache/skywalking-go/actions/runs/7913023188/job/21600110686#step:5:1423), the order of consuming messages is not in the order of message sending, this will affect the stability of plugins test, this time I adjusted the relevant files again

test/plugins/scenarios/rocketmq/config/excepted.yml
CodePrometheus: fixed

mrproliu (approved):

PR #170, opened 2024-02-14
feature: add golang 1.21, 1.22 compatibility for docker image

darknos:
Currently SkyWalking Go provided image as the base image, which currently supports the following Go versions: 1.16, 1.17, 1.18, 1.19, 1.20

but we need to support 1.21 and even 1.22


darknos:
I see some checks failed but they are not connected to the PR... :(

mrproliu:
I have restart the CI, let's wait.

PR #173, opened 2024-02-26
feature: add support for Pulsar

CodePrometheus:
### summary
[Pulsar](https://github.com/apache/pulsar-client-go) is a native go client maintained by the apache pulsar team.

- [x] Update the docs.
- [x] Pass CI test.
- [ ] Pass Plugin test.

### test report
image


CodePrometheus:
Ref https://github.com/apache/skywalking/pull/11937. 

Maybe I need some time, please hold for a while, thanks. 

AFAIK, We have Pulsar monitoring and agent plugins (Java is ready and Go are currently in progress) for Pulsar, skywalking-showcase needs to be added accordingly? I am willing to try.

wu-sheng:
Showcase is fine, it can't hold all plugin examples there.

wu-sheng:
Meanwhile, you should try and test service hierarchy for Pulsar, if you want. especially from virtual MQ to MQ server.

CodePrometheus:
> Meanwhile, you should try and test service hierarchy for Pulsar, if you want. especially from virtual MQ to MQ server.

Copy that, I want to try it.

CodePrometheus:
PTAL.

plugins/pulsar/pulsar/receive_consumer.go
```diff
@@ -0,0 +1,70 @@
...
+	topic := nativeConsumer.options.Topic
+	lookup, err := nativeConsumer.client.lookupService.Lookup(topic)
+	if err != nil {
+		return err
+	}
+	message := result[0].(pulsar.Message)
+	peer := lookup.PhysicalAddr.String()
+	operationName := pulsarReceivePrefix + topic + pulsarReceiveSuffix
+
+	span, err := tracing.CreateEntrySpan(operationName, func(headerKey string) (string, error) {
```
mrproliu, line 51: Why we don't move `CreateEntrySpan` to the `BeforeInvoke`? It looks like we created a span and stopped them quickly.

test/plugins/scenarios/pulsar/config/excepted.yml
```diff
@@ -0,0 +1,142 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+segmentItems:
+  - serviceName: pulsar
+    segmentSize: 4
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: GET:/health
```
mrproliu, line 23: Please delete the `GET:/health` checker.

test/plugins/scenarios/pulsar/config/excepted.yml
```diff
@@ -0,0 +1,142 @@
...
+            isError: false
+            spanType: Exit
+            peer: pulsar://pulsar-server:6650
+            skipAnalysis: false
+            tags:
+              - { key: mq.broker, value: not null }
+              - { key: mq.topic, value: not null }
+              - { key: mq.msg.id, value: not null }
+          - operationName: Pulsar/sw-topic-1/Consumer
+            parentSpanId: 0
```
mrproliu, line 56: Why the consumer parent span id is `0`(span `GET:/execute`), I think It should be `Pulsar/sw-topic-1/Producer`. Am I right?

test/plugins/scenarios/pulsar/config/excepted.yml
```diff
@@ -0,0 +1,142 @@
...
+            spanType: Entry
+            peer: ''
+            skipAnalysis: false
+            tags:
+              - { key: http.method, value: GET }
+              - { key: url, value: 'service:8080/execute' }
+              - { key: status_code, value: '200' }
+      - segmentId: not null
+        spans:
+          - operationName: Pulsar/sw-topic-2/Producer/Callback
```
mrproliu, line 105: When will we have the `Callback` span, It's a local span and not reference? Could you please add the screenshot about this plugin test? I'm not sure what's the real show like.

plugins/pulsar/pulsar/receive_consumer.go
CodePrometheus: Because the `Extractor` function requires the `message` in the `result[0]` to inject the `headerKey`.

test/plugins/scenarios/pulsar/config/excepted.yml
CodePrometheus: Now as shown in the figure, when the two test methods are executed separately, it's as expected.
image

mrproliu: Then, why does the async producer have the children consumer, but the synchronized producer doesn't?

test/plugins/scenarios/pulsar/config/excepted.yml
CodePrometheus: Done.

test/plugins/scenarios/pulsar/config/excepted.yml
CodePrometheus: I think it's a problem with the test code, and I think I've found the solution, please take a look again, thx.

test/plugins/scenarios/pulsar/config/excepted.yml
CodePrometheus: In the current test code, both consumer and producer are triggered by `http`.

test/plugins/scenarios/pulsar/config/excepted.yml
CodePrometheus: I noticed that I have also retained this spans in other plugin's file, I'll delete it from future PR.

CodePrometheus:
`go: /skywalking-go/go.work:3: unknown directive: toolchain` 
Very strange, I tried it locally and it was passed.
image

mrproliu:
Could you take a look on the go.work file under the root directory? Maybe change this file in a mistake?

mrproliu (approved):

PR #174, opened 2024-03-10
[Test] Update the supported versions for fiber

CodePrometheus:
Update the supported versions for fiber.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@ Changes by Version
 ==================
 Release Notes.
 
+0.5.0
+------------------
+#### Plugins
+* Update the supported versions for fiber.
+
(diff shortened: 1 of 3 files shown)


wu-sheng:
49.0 is not supposed?

CodePrometheus:
I suspect that this https://github.com/golang/go/issues/62409 may be causing the error in the pulsar CI test.

CodePrometheus:
It seems that the env may be unstable, could we discuss it in Slack?

wu-sheng:
Yes, slack is open for discussion. @mrproliu may be back tomorrow or today night.

CodePrometheus:
Just to locate the problem, I close it, thx.

PR #176, opened 2024-03-16
feature: add support for segmentio-kafka

CodePrometheus:
### Summary

[segmentio-kafka](https://github.com/segmentio/kafka-go) is a lightweight Kafka library in Go.

- [x] Update the docs.
- [x] Pass CI test.
- [x] Pass Plugin test.

### Test Report

image


CodePrometheus:
Ref https://github.com/apache/skywalking/pull/12027

Superskyyy:
Please fix lint and stuff, thanks

test/plugins/scenarios/segmentio-kafka/main.go
```diff
@@ -0,0 +1,149 @@
...
+	for _, test := range tests {
+		fmt.Printf("excute test case: %s\n", test.name)
+		if subErr := test.fn(); subErr != nil {
+			fmt.Printf("test case %s failed: %v", test.name, subErr)
+		}
+	}
+}
+
+func simpleMsg() error {
+	consumerHelper()
```
mrproliu, line 82: Can we move the consumer into the goroutine when starting main rather than in the HTTP handler, I think it more like the real scenario.

CodePrometheus: In the test class, if the consumer is started too early, the consumer may be in front of the producer when creating the span.

CodePrometheus:
~~~
ALLOW_PLAINTEXT_LISTENER: yes

services.kafka-server.environment.ALLOW_PLAINTEXT_LISTENER contains true, which is an invalid type, it should be a string, number, or a null
~~~
It seems that docker-compose v2 has been verified. I'll correct it.

test/plugins/scenarios/segmentio-kafka/main.go
wu-sheng: I am not following why consumer starts first matter the test.
It is a common real world scenario. 
Could you explain more?

CodePrometheus: Since I want the overall display to flow to http->produce->consume and be displayed in a segmentId, so I made some control logic for the test code.
And I found some examples that confuse me a bit, [kafka-java](https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/kafka-scenario/config/expectedData.yaml) and [activemq-java](https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/activemq-artemis-2.x-scenario/config/expectedData.yaml), the order here is sorted by the time span was collected, am I understanding it correctly?

wu-sheng: Spans are sorted by the hierarchy relationship, rather than time. When the consumer comes first, it should wait for the message and no consuming happens.

CodePrometheus: Get it, I'll adjust it.

CodePrometheus:
Updated.

test/plugins/scenarios/segmentio-kafka/config/excepted.yml
```diff
@@ -0,0 +1,74 @@
...
+            componentId: 41
+            isError: false
+            spanType: Entry
+            peer: kafka-server:9092
+            skipAnalysis: false
+            tags:
+              - { key: mq.broker, value: not null }
+              - { key: mq.topic, value: not null }
+            refs:
+              - { parentEndpoint: 'GET:/execute', networkAddress: ' kafka-server:9092',
```
mrproliu, line 69: Why there have a blank prefix in the network address? Please fix it.

CodePrometheus: A slip of the thumb, fixed.

mrproliu (approved):

CodePrometheus:
~~~
dependency failed to start: container gin-service-1 exited (127)
WORK=/tmp/go-build4103001024
# text/template/parse
runtime: newstack sp=0xc00004ff90 stack=[0xc000d[700](https://github.com/apache/skywalking-go/actions/runs/8313025290/job/22748663526?pr=176#step:7:701)00, 0xc000d78000]
	morebuf={pc:0x7ffd347e6bb2 sp:0xc00004ffa0 lr:0x0}
	sched={pc:0x44b80b sp:0xc00004ff98 lr:0x0 ctxt:0x0}
runtime: gp=0xc0000bd6c0, goid=39, gp->status=0x2
 runtime: split stack overflow: 0xc00004ff90 < 0xc000d70000
fatal error: runtime: split stack overflow
~~~

BTW, do we need to test plugin separately on macOS?

PR #177, opened 2024-03-20
Add support trace ignore

CodePrometheus:
### Summary

If the operation name of the first span is matching, this segment should be ignored.(multiple split by ","), matching rules follow Ant Path match style, like /path/*, /path/**, /path/?.

### DIFF
--- a/docs/en/agent/tracing-metrics-logging.md
+++ b/docs/en/agent/tracing-metrics-logging.md
@@ -22,11 +22,11 @@ If you wish to disable a particular plugin to prevent enhancements related to th
 
 The basic configuration is as follows:
 
-| Name                    | Environment Key            | Default Value                                                | Description                                                                                                              |
-|-------------------------|----------------------------|--------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------|
-| agent.sampler           | SW_AGENT_SAMPLER           | 1                                                            | Sampling rate of tracing data, which is a floating-point value that must be between 0 and 1.                             |
-| agent.ignore_suffix     | SW_AGENT_IGNORE_SUFFIX     | .jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg | If the operation name of the first span is included in this set, this segment should be ignored.(multiple split by ","). |
-| agent.trace_ignore_path | SW_AGENT_TRACE_IGNORE_PATH |                                                              | If the operation name of the first span is matching, this segment should be ignored.(multiple split by ",").             |
+| Name                    | Environment Key            | Default Value                                                | Description                                                                                                                                          |
+|-------------------------|----------------------------|--------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------|
+| agent.sampler           | SW_AGENT_SAMPLER           | 1                                                            | Sampling rate of tracing data, which is a floating-point value that must be between 0 and 1.                                                         |
+| agent.ignore_suffix     | SW_AGENT_IGNORE_SUFFIX     | .jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg | If the suffix obtained by splitting the operation name by the last index of "." in this set, this segment should be ignored.(multiple split by ","). |
+| agent.trace_ignore_path | SW_AGENT_TRACE_IGNORE_PATH |                                                              | If the operation name of the first span is matching, this segment should be ignored.(multiple split by ",").                                         |
 
 ## Metrics
 
--- a/docs/en/agent/tracing-metrics-logging.md
+++ b/docs/en/agent/tracing-metrics-logging.md
@@ -22,10 +22,11 @@ If you wish to disable a particular plugin to prevent enhancements related to th
 
 The basic configuration is as follows:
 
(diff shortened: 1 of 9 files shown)


wu-sheng:
> matching rules follow Ant Path match style, like /path/*, /path/**, /path/?.

Is this a question or something?

docs/en/agent/tracing-metrics-logging.md
```diff
@@ -22,10 +22,11 @@ If you wish to disable a particular plugin to prevent enhancements related to th
...
+|-------------------------|----------------------------|--------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------|
+| agent.sampler           | SW_AGENT_SAMPLER           | 1                                                            | Sampling rate of tracing data, which is a floating-point value that must be between 0 and 1.                             |
+| agent.ignore_suffix     | SW_AGENT_IGNORE_SUFFIX     | .jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg | If the operation name of the first span is included in this set, this segment should be ignored.(multiple split by ","). |
+| agent.trace_ignore_path | SW_AGENT_TRACE_IGNORE_PATH |                                                              | If the operation name of the first span is matching, this segment should be ignored.(multiple split by ",").             |
```
wu-sheng, line 29: If the entry span(first span) of the segment is `GET:/ABC`, in theory, when I set `agent.ignore_suffix=GET:/ABC`, what will happen?

plugins/core/tracer_ignore.go
```diff
@@ -0,0 +1,144 @@
...
+	for _, suffix := range ignoreSuffix {
+		if suffix == operationName[suffixIdx:] {
+			return true
+		}
+	}
+	return false
+}
+
+func traceIgnorePath(operationName string, ignorePath []string) bool {
+	op := strings.Index(operationName, COLON)
```
mrproliu, line 47: Why need to check the operation name must-have :, if the operation is generated by another RPC framework(not HTTP), then this configuration will be broken.

tools/go-agent/config/agent.default.yaml
```diff
@@ -30,8 +30,11 @@ agent:
   correlation:
     max_key_count: ${SW_AGENT_CORRELATION_MAX_KEY_COUNT:3}
     max_value_size: ${SW_AGENT_CORRELATION_MAX_VALUE_SIZE:128}
-  # If the operation name of the first span is included in this set, this segment should be ignored.(multiple split by ",")
+  # If the operation name of the first span is included in this set, this segment should be ignored.(multiple split by ",").
   ignore_suffix: ${SW_AGENT_IGNORE_SUFFIX:.jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg}
+  # If the operation name of the first span is matching, this segment should be ignored.(multiple split by ",").
+  # Matching rules follow Ant Path match style, like /path/*, /path/**, /path/?.
```
mrproliu, line 36: Please add more comments about when the user should use `*`, `**`, or `?`.

docs/en/agent/tracing-metrics-logging.md
CodePrometheus: `ignore_suffix` is currently split according to `.` to get the suffix for matching, so `GET:/ABC` will not be filtered.

plugins/core/tracer_ignore.go
CodePrometheus: Are you referring to the occurrence of an empty value for operationName? I'm currently conducting validation based on the operationName dimension.

docs/en/agent/tracing-metrics-logging.md
wu-sheng: I see. Could you add this to the doc? Because in the Java agent, it is using `endOf` only, so, actually, it could use for complete match filter, although we don't recommend it.

plugins/core/tracer_ignore.go
mrproliu: Nope, I mean why the operation name must-have `:`? Not all operation names contain this symbol. It could make users confused as to why It does not work for other span names.

CodePrometheus: I did overlook this issue and was only considering web requests before, I will adjust my strategy.

docs/en/agent/tracing-metrics-logging.md
CodePrometheus: Done.

tools/go-agent/config/agent.default.yaml
CodePrometheus: Done.

plugins/core/tracer_ignore.go
```diff
@@ -0,0 +1,133 @@
...
+func traceIgnorePath(operationName string, ignorePath []string) bool {
+	for _, pattern := range ignorePath {
+		if normalMatch(pattern, 0, operationName, 0) {
+			return true
+		}
+	}
+	return false
+}
+
+func normalMatch(pat string, p int, str string, s int) bool {
```
mrproliu, line 50: Can you add more comments to the parameters? This can help developers better understand your code.

CodePrometheus: Done.

mrproliu (approved):

PR #178, opened 2024-03-20
Supporting http headers collection for Gin

IceSoda177:
Supporting http headers collection for Gin

### DIFF
--- a/plugins/gin/config.go
+++ b/plugins/gin/config.go
@@ -15,15 +15,10 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package operator
+package gin
 
-type ToolsOperator interface {
-	ReflectGetValue(instance interface{}, filters []interface{}) interface{}
-	GetEnvValue(key string) string
-	ParseFloat(val string, bitSize int) (float64, error)
-	ParseBool(val string) bool
-	ParseInt(val string, base, bitSize int) (int64, error)
-	ParseStringArray(val string) ([]string, error)
-	Atoi(val string) (int, error)
-	NewSyncMap() interface{}
+//skywalking:config gin
+var config struct {
+	CollectRequestHeaders []string `config:"collect_request_headers"`
+	HeaderLengthThreshold int      `config:"header_length_threshold"`
 }
--- a/plugins/gin/intercepter.go
+++ b/plugins/gin/intercepter.go
@@ -19,11 +19,13 @@ package gin
 
 import (
 	"fmt"
-
-	"github.com/gin-gonic/gin"
+	"net/http"
+	"strings"
 
 	"github.com/apache/skywalking-go/plugins/core/operator"
 	"github.com/apache/skywalking-go/plugins/core/tracing"
+
+	"github.com/gin-gonic/gin"
 )
 
 type ContextInterceptor struct {
@@ -45,6 +47,11 @@ func (h *ContextInterceptor) BeforeInvoke(invocation operator.Invocation) error 
 	if err != nil {
 		return err
 	}
+
+	if len(config.CollectRequestHeaders) > 0 {
+		collectRequestHeaders(s, context.Request.Header)
+	}
+
 	invocation.SetContext(s)
 	return nil
 }
@@ -69,4 +76,26 @@ func isFirstHandle(c interface{}) bool {
 		return context.index < 0
 	}
 	return true
+}
+
+func collectRequestHeaders(span tracing.Span, requestHeaders http.Header) {
+	var headerTagValues []string
+	for _, header := range config.CollectRequestHeaders {
+		var headerValue = requestHeaders.Get(header)
+		if headerValue != "" {
+			headerTagValues = append(headerTagValues, header+"="+headerValue)
+		}
+	}
+
+	if len(headerTagValues) == 0 {
+		return
+	}
+
+	tagValue := strings.Join(headerTagValues, "\n")
+	if len(tagValue) > config.HeaderLengthThreshold {
+		maxLen := config.HeaderLengthThreshold
+		tagValue = tagValue[:maxLen]
+	}
+
+	span.Tag(tracing.TagHTTPHeaders, tagValue)
 }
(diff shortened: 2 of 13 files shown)


plugins/gin/intercepter.go
```diff
@@ -70,3 +77,26 @@ func isFirstHandle(c interface{}) bool {
 	}
 	return true
 }
+
+func collectHttpHeaders(span tracing.Span, requestHeaders http.Header) {
+	if len(runtimeConfig.CollectHttpHeaders) == 0 {
```
wu-sheng, line 82: If len==0 then split? Do you mis-write this?

wu-sheng: You should update the gin plugin test to verify this feature.

wu-sheng:
@CodePrometheus Would you like to review this?

plugins/gin/intercepter.go
IceSoda177: > If len==0 then split? Do you mis-write this?

They are different config. I use runtimeConfig to store header list, cause the config can't accept array type config param.

IceSoda177: > You should update the gin plugin test to verify this feature.

right

tools/go-agent/config/agent.default.yaml
```diff
@@ -97,3 +97,5 @@ plugin:
     redis:
       # Limit the bytes size of redis args request
       max_args_bytes: ${SW_AGENT_PLUGIN_CONFIG_REDIS_MAX_ARGS_BYTES:1024}
+    gin: 
+      collect_http_headers: ${SW_AGENT_PLUGIN_CONFIG_GIN_COLLECT_HTTP_HEADERS:}
```
mrproliu, line 101: I think you missing the configuration declaration. please take a look at other plugins, such as https://github.com/apache/skywalking-go/blob/main/plugins/go-redisv9/config.go

mrproliu: And the default value should be `false` I think.

IceSoda177: Yes, config.go is missing

wu-sheng:
UT is just a small thing, it is not very useful in agent mode. We have specific plugin test, https://skywalking.apache.org/docs/skywalking-go/next/en/development-and-contribution/write-plugin-testing/

The existing one is at here, https://github.com/apache/skywalking-go/tree/main/test/plugins/scenarios/gin

You should enhance it, verify locally, and update.

CodePrometheus:
> @CodePrometheus Would you like to review this?

I am more than willing to participate in code review, and I will pay attention during my free time at work~ (Beijing time)

CodePrometheus:
@IceSoda177 hello bro, could you help to update the docs/en/agent/plugin-configurations.md? And plz fix the CI.

plugins/gin/config.go
```diff
@@ -0,0 +1,23 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package gin
+
+//skywalking:config gin
+var config struct {
+	CollectHttpHeaders string `config:"collect_http_headers"`
```
mrproliu, line 22: A small suggestion, could you change his data type to `[]string`? Although it may be a bit challenging, you might need to provide parsing capabilities in the agent core, so that you don't have to use `runtimeConfig` and reprocess it for each request. If it's still difficult for you, I can help with it. :)

IceSoda177:
> UT is just a small thing, it is not very useful in agent mode. We have specific plugin test, https://skywalking.apache.org/docs/skywalking-go/next/en/development-and-contribution/write-plugin-testing/
> 
> The existing one is at here, https://github.com/apache/skywalking-go/tree/main/test/plugins/scenarios/gin
> 
> You should enhance it, verify locally, and update.

UT is no doubt necessary.

IceSoda177:
> @IceSoda177 hello bro, could you help to update the docs/en/agent/plugin-configurations.md? And plz fix the CI.

Will do

plugins/gin/config.go
IceSoda177: Yes, that is bothering me. And I want to solve that. where should I getting start?

wu-sheng:
> > UT is just a small thing, it is not very useful in agent mode. We have specific plugin test, https://skywalking.apache.org/docs/skywalking-go/next/en/development-and-contribution/write-plugin-testing/
> > The existing one is at here, https://github.com/apache/skywalking-go/tree/main/test/plugins/scenarios/gin
> > You should enhance it, verify locally, and update.
> 
> UT is no doubt necessary.

Yes. But it is not verifying the real case. It only verified expected parameter works.

plugins/gin/config.go
mrproliu: You need to follow these steps:
1. Adding the `[]string` case in there: https://github.com/apache/skywalking-go/blob/main/tools/go-agent/instrument/plugins/enhance_config.go#L219-L240
2. Adding `ParseStringArray` into the [ToolOperator interface](https://github.com/apache/skywalking-go/blob/main/plugins/core/operator/tools.go) and [implement it](https://github.com/apache/skywalking-go/blob/main/plugins/core/tracer_tools.go).

IceSoda177: Working on it.

IceSoda177:
> UT is just a small thing, it is not very useful in agent mode. We have specific plugin test, https://skywalking.apache.org/docs/skywalking-go/next/en/development-and-contribution/write-plugin-testing/
> 
> The existing one is at here, https://github.com/apache/skywalking-go/tree/main/test/plugins/scenarios/gin
> 
> You should enhance it, verify locally, and update.

The wall is making me crazy...

wu-sheng:
There is nothing we can do about your network environment. And it is common for go dev, we don't add anything new.

IceSoda177:
All modification and test is done, any things else do I need to do?

wu-sheng:
You still miss this,

> The existing one is at here, https://github.com/apache/skywalking-go/tree/main/test/plugins/scenarios/gin
You should enhance it, verify locally, and update.

We need this feature verified in this demo application instrumentation.

IceSoda177:
> all thing, it is not very useful in agent mode. We have specif

Sorry, I get your point now! I will update the scenario later.
My network problem finally solved by using vpn. Otherwise Docker images from google can not be downloaded, and downloading go dependents from github in container will failled.

wu-sheng:
Take your time, it is urgent.

mrproliu (approved):
LGTM, Once the CI is success, I think It could be merged. :)

PR #185, opened 2024-05-27
feat: enhance version info

ShyunnY:
Enhance command version information output.

**Before**:
```shell
$ ./skywalking-go-agent-0.4.0-windows-amd64 -version
skywalking-go agent version: v0.4.0
```

**After**:
```shell
$ ./skywalking-go-agent-v0.4.0-windows-amd64 -version
skywalking-go agent
 Version: v0.4.0
 Go version: 1.22.2
 Git commit: 7948b43
```
> _NOTE: Added one space indentation (more elegant than two spaces) to make it more hierarchical_


VERSION
```diff
@@ -0,0 +1 @@
+v0.4.0
```
mrproliu, line 1: When this file will be update? Can we update it in the release flow or something else?

ShyunnY:
I'm sorry that a file (`helper_test.go`) that does not belong to this PR has been submitted. I will solve it.

VERSION
ShyunnY: Yes, we should update this file every time a new version is released

mrproliu: Then, please read the release documentation, I think we don't need to setting this value manually. Can we just using environment rather than release file about this value?

ShyunnY: I checked `release.sh`, from which I can see that the version is controlled through envs. We can compare these two methods:

**use files:**
+ Version information is visible in the version control system, and the change history of the version information can be tracked, including when, by whom and why changes were made.
+ Version information is clearly visible in the project code, making it easy to understand and find.
+ For team members who need to frequently view or modify version information, using files may be more intuitive.

**use env:**
+ Can set different versions for each environment or build without modifying the code.
+ For projects using continuous integration/continuous deployment (CI/CD) tools, environment variables make it easy to set the correct version when building or deploying.

I'm voting to use file and I'd like to hear your and everyone's thoughts.

mrproliu: We can use the release page on Github to see all the content you need to view in the file.
If we are using file mode, we still need to modify this file before release, It increases the complexity of maintaining releases. Unless you make the file automatically generated when building the project.

mrproliu: Additionally, regardless of the final storage method used, please manage all version-related contents(agent version, Go version number, and Git commit version number) in the same way. Otherwise, other developers or maintenance personnel may become confused.

ShyunnY: > Additionally, regardless of the final storage method used, please manage all version-related contents(agent version, Go version number, and Git commit version number) in the same way. Otherwise, other developers or maintenance personnel may become confused.

In fact: version, goVersion, and gitCommit records are all determined when the version is released.

ShyunnY: > We can use the release page on Github to see all the content you need to view in the file. If we are using file mode, we still need to modify this file before release, It increases the complexity of maintaining releases. Unless you make the file automatically generated when building the project.

That seems to make sense and I'll think about it. 
For now, let’s use the env way first!    :)

tools/go-agent/Makefile
```diff
@@ -23,12 +23,15 @@ BINARY = skywalking-go-agent
...
+GIT_COMMIT := $(shell $(GIT) rev-parse --short HEAD)
+GO_VERSION := $(shell $(GO) version | sed 's/go version go\([0-9.]*\).*/\1/')
+VERSION_PACKAGE := main
 PROTOC = protoc
 GO_PATH = $$($(GO) env GOPATH)
 GO_BUILD = $(GO) build
 GO_LINT = $(GO_PATH)/bin/golangci-lint
 GO_BUILD_FLAGS = -v
-GO_BUILD_LDFLAGS = -X main.version=$(VERSION)
+GO_BUILD_LDFLAGS = -X $(VERSION_PACKAGE).version=$(VERSION) -X $(VERSION_PACKAGE).goVersion=$(GO_VERSION) -X $(VERSION_PACKAGE).gitCommit=$(GIT_COMMIT)
```
wu-sheng, line 34: I have concern of doing this. Note, in the ASF, the release tar means the one on the SVN. Not the github one.

ShyunnY: Don't worry, this only captures environment information during compilation and will not be interfered by other factors

wu-sheng: I am not saying this could cause an error. I am saying this would generate a different binary tar when users build from the official release. This should not happen.

All builds of official releases and binaries should be reproducible.

ShyunnY: Sorry, I misunderstood you.

When users refer to [build-docs](https://skywalking.apache.org/docs/skywalking-go/latest/en/development-and-contribution/build-and-use-agent/#build-the-agent) to build, in fact The obtained version is also Unknown
```shell
$ cd skywalking-go && make build
...

$ cd ./bin/ && ./skywalking-go-agent--windows-amd64 -version
skywalking-go agent version: unknown
```
In sw-go,`release.sh` actually controls the version:
https://github.com/apache/skywalking-go/blob/7948b433093a6ff51a7131a8f86ecfe5fd9ed56c/tools/release/create_bin_release.sh#L26-L38

Of course, I also support this after revision.  If the user uses `make build` directly, the output will be as follows:
```shell
$ cd skywalking-go && make build
...

$ cd ./bin/ && ./skywalking-go-agent--windows-amd64 -version
skywalking-go agent
 Version: Unknown
 Go version: go1.22.2
 Git commit: a6824f6
```

I'm not sure if I understand you correctly, thanks for your reply

wu-sheng: Then it was not correct.
The version and commit ID should be a part of source tar, to make sure it could be repackaged from the source.

wu-sheng: @mrproliu I am not sure what the way of current go agent. But the policy should be, if we have version and commit ID in the binary in the release we built, users should have that when they build from our release(source tar in ASF).

mrproliu: Right. I have talked with @ShyunnY, He will save the version information as a file and save it to the release package. Then, we can resolve this issue.

Makefile
```diff
@@ -29,11 +29,13 @@ GO_TEST_LDFLAGS =
...
 LINT_FILE_PATH = $(REPODIR).golangci.yml
 
+VERSION_FILE=VERSION
+
 SHELL = /bin/bash
 
 HUB ?= docker.io/apache
 PROJECT ?= skywalking-go
-VERSION ?= $(shell git rev-parse --short HEAD)
+GIT_VERSION ?= $(shell git rev-parse --short HEAD)
```
mrproliu, line 38: You still have not fixed this issue, In the released source tar, there will not be a GitHub, so it will fail.

ShyunnY: Can we build the corresponding VERSION file in `create_source_release.sh` and package it into the src package?

Even if the user downloads and compiles through the src package, the correct VERSION file content can be read

mrproliu: Yes. When releasing, the version information needs to be written into the release source tar so that users downloading the release tar hosted by Apache will already have the version information included.

tools/go-agent/Makefile
ShyunnY: I insert the VERSION file when compiling src tgz, so that the information obtained by users directly downloading the source package is always consistent with the information obtained by git clone. 

This presents the idea that "no matter how you install, the information is always consistent"

mrproliu:
Could you also adding the version file to the gitignore file, encase people add them to this repo.

tools/go-agent/cmd/helper.go
```diff
@@ -45,16 +49,28 @@ func PrintUsageWithExit() {
...
+	}
+	if err := versionTmpl.Execute(os.Stdout, versionInfo); err != nil {
+		fmt.Fprintln(os.Stdout, err)
 	}
-	fmt.Printf("skywalking-go agent version: %s\n", res)
+	os.Exit(1)
 }
 
-var usage = `
+var versionTemplate = `
```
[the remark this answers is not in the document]
ShyunnY: Done!

mrproliu (approved):

PR #187, opened 2024-06-18
Resolve panic when root span finish

mrproliu:


CHANGES.md
```diff
@@ -13,6 +13,10 @@ Release Notes.
 * Support http headers collection for Gin
 * Support higher versions of grpc.
 
+### Bug Fixes
+* Fix panic error when root span finished.
+* Update the peer address is not required when creating exit span.
```
wu-sheng, line 18: If peer is absent, the error will occur at OAP end.

mrproliu: I see the OAP side if the peer is absent(blank string), then I would have a strategy on this. It seems that sometimes the peer address cannot be obtained.

wu-sheng: If it can't be obtained, we should consider this is a plugin bug. 
OAP failove mechanism is just for avoiding errors, but logically it is still wrong.

This causes(for exit span)
1. No virtual service(DB, MQ, etcd.)
1. Trace broken at server side, as the trace context in sw8 header is wrong.
1. Topology broken in case <2>

In theory, when peer is absent, we should create an exit span.

mrproliu: I have updated the error message when creating an exit span with an empty peer address. Then the exit span will not created.

wu-sheng (approved):

Qiangel:
I use redis.ClusterClient in go-redisv9. Because of the AfterInvoke of intercepter.go, redis cluster uses c.AddHook(newRedisHook("")), resulting in an error message indicating that the following peer is empty. Is this experiencing the same problem?
There are also other issues with go-redisv9, when I set the environment variable sample rate not to 1, eg `export SW_AGENT_SAMPLE="0.1"`, in the 'trace.go' 'createSpan0' method, 'newNoopSpan()' is returned, causing 'CreateExitSpan' to return 'errors.New("span type is wrong")'.
This will pick up the error in ClusterClient, causing the redis business code to fail to judge correctly.

Qiangel:
I use redis.ClusterClient in go-redisv9. Because of the AfterInvoke of intercepter.go, redis cluster uses `c.AddHook(newRedisHook(""))`, resulting in an error message indicating that the following peer is empty. Is this experiencing the same problem?
There are also other issues with go-redisv9, when I set the environment variable sample rate not to 1, eg `export SW_AGENT_SAMPLE="0.1"`, in the `trace.go` `createSpan0` method, `newNoopSpan()` is returned, causing `CreateExitSpan` to return `errors.New("span type is wrong")`.
This will pick up the error in ClusterClient, causing the redis business code to fail to judge correctly.

PR #188, opened 2024-07-12
feat: impl http.Hijacker

fzdwx:
Work with weboskcet library(e.g https://github.com/gorilla/websocket), `writerWrapper` should implement `http.Hijacker`

### DIFF
--- a/plugins/mux/serve_interceptor.go
+++ b/plugins/mux/serve_interceptor.go
@@ -90,13 +90,12 @@ func (w *writerWrapper) WriteHeader(statusCode int) {
 
 func newWriterWrapperWithHijacker(writer http.ResponseWriter, hijacker http.Hijacker) *writerWrapperWithHijacker {
 	return &writerWrapperWithHijacker{
-		w:        newWriterWrapper(writer),
-		Hijacker: hijacker,
+		ResponseWriter: newWriterWrapper(writer),
+		Hijacker:       hijacker,
 	}
 }
 
 type writerWrapperWithHijacker struct {
-	w *writerWrapper
 	http.ResponseWriter
 	http.Hijacker
 }
--- a/plugins/mux/serve_interceptor.go
+++ b/plugins/mux/serve_interceptor.go
@@ -51,8 +51,8 @@ func (n *ServeHTTPInterceptor) AfterInvoke(invocation operator.Invocation, resul
 		return nil
 	}
 	span := invocation.GetContext().(tracing.Span)
-	if wrapped, ok := invocation.Args()[0].(StatusCodeHolder); ok {
-		span.Tag(tracing.TagStatusCode, fmt.Sprintf("%d", wrapped.GetStatusCode()))
+	if wrapped, ok := invocation.Args()[0].(*writerWrapper); ok {
+		span.Tag(tracing.TagStatusCode, fmt.Sprintf("%d", wrapped.statusCode))
 	}
 	span.End()
 	return nil
@@ -70,20 +70,11 @@ func newResponseWriter(val interface{}) http.ResponseWriter {
 	return rw
 }
 
-func newWriterWrapper(writer http.ResponseWriter) WriterWrapper {
+func newWriterWrapper(writer http.ResponseWriter) *writerWrapper {
 	return &writerWrapper{
 		ResponseWriter: writer,
 		statusCode:     http.StatusOK,
 	}
-}
-
-type StatusCodeHolder interface {
-	GetStatusCode() int
-}
-
-type WriterWrapper interface {
-	http.ResponseWriter
-	StatusCodeHolder
 }
 
 type writerWrapper struct {
@@ -97,18 +88,14 @@ func (w *writerWrapper) WriteHeader(statusCode int) {
 	w.ResponseWriter.WriteHeader(statusCode)
 }
 
-func (w *writerWrapper) GetStatusCode() int {
-	return w.statusCode
-}
-
 func newWriterWrapperWithHijacker(writer http.ResponseWriter, hijacker http.Hijacker) *writerWrapperWithHijacker {
 	return &writerWrapperWithHijacker{
-		WriterWrapper: newWriterWrapper(writer),
-		Hijacker:      hijacker,
+		ResponseWriter: newWriterWrapper(writer),
+		Hijacker:       hijacker,
 	}
 }
 
 type writerWrapperWithHijacker struct {
-	WriterWrapper
+	http.ResponseWriter
 	http.Hijacker
 }
--- a/plugins/mux/serve_interceptor.go
+++ b/plugins/mux/serve_interceptor.go
@@ -63,7 +63,7 @@ func newWriterWrapper(rw interface{}) *writerWrapper {
 	writer := rw.(http.ResponseWriter)
 	hijacker, ok := rw.(http.Hijacker)
 	if !ok {
-		log.Errorf("http.ResponseWriter does not implement http.Hijacker")
+		log.Warnf("http.ResponseWriter does not implement http.Hijacker")
 	}
 
 	return &writerWrapper{
--- a/plugins/mux/serve_interceptor.go
+++ b/plugins/mux/serve_interceptor.go
@@ -19,6 +19,7 @@ package mux
 
 import (
 	"fmt"
+	"github.com/apache/skywalking-go/plugins/core/log"
 	"net/http"
 
 	"github.com/apache/skywalking-go/plugins/core/operator"
@@ -60,10 +61,14 @@ func (n *ServeHTTPInterceptor) AfterInvoke(invocation operator.Invocation, resul
 
 func newWriterWrapper(rw interface{}) *writerWrapper {
 	writer := rw.(http.ResponseWriter)
-	controller := http.NewResponseController(writer)
+	hijacker, ok := rw.(http.Hijacker)
+	if !ok {
+		log.Errorf("http.ResponseWriter does not implement http.Hijacker")
+	}
+
 	return &writerWrapper{
 		ResponseWriter: writer,
-		Hijacker:       controller,
+		Hijacker:       hijacker,
 		statusCode:     http.StatusOK,
 	}
 }
--- a/plugins/mux/serve_interceptor.go
+++ b/plugins/mux/serve_interceptor.go
@@ -40,7 +40,11 @@ func (n *ServeHTTPInterceptor) BeforeInvoke(invocation operator.Invocation) erro
 		return err
 	}
 
-	invocation.ChangeArg(0, newWriterWrapper(invocation.Args()[0]))
+	rw, err := newWriterWrapper(invocation.Args()[0])
+	if err != nil {
+		return err
+	}
+	invocation.ChangeArg(0, rw)
 	invocation.SetContext(s)
 	return nil
 }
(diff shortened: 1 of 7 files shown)


plugins/mux/serve_interceptor.go
```diff
@@ -58,8 +57,19 @@ func (n *ServeHTTPInterceptor) AfterInvoke(invocation operator.Invocation, resul
 	return nil
 }
 
+func newWriterWrapper(rw interface{}) *writerWrapper {
+	writer := rw.(http.ResponseWriter)
+	hijacker := rw.(http.Hijacker)
```
mrproliu, line 62: Is any possible the agent will got panic when cast failure? Please do more explain on this.

mrproliu (commented):
Is this changes related to the plugin test? If so, please update the test case.

plugins/mux/serve_interceptor.go
fzdwx: > Is any possible the agent will got panic when cast failure? Please do more explain on this.

If it were the original `http.ResponseWirter` then the panic would definitely not happen. 

https://github.com/golang/go/blob/master/src/net/http/server.go#L2155

fzdwx:
> Is this changes related to the plugin test? If so, please update the test case.

I'm getting familiar with how to write test cases.

plugins/mux/serve_interceptor.go
mrproliu: Then, please added a cast check please, some third-party library may change this.

fzdwx: sure.

plugins/mux/serve_interceptor.go
```diff
@@ -58,8 +61,23 @@ func (n *ServeHTTPInterceptor) AfterInvoke(invocation operator.Invocation, resul
 	return nil
 }
 
+func newWriterWrapper(rw interface{}) (*writerWrapper, error) {
+	writer := rw.(http.ResponseWriter)
+	hijacker, ok := rw.(http.Hijacker)
+	if !ok {
+		return nil, fmt.Errorf("http.ResponseWriter does not implement http.Hijacker")
```
mrproliu, line 68: If cast failure, could you align with the older strategy? We should not break other framework use.

fzdwx: Set hijacker to nil? Or implement the `Hijack` method ？

mrproliu: I think the original wrapper has already implement the hijick interface? Right? I don't know why you add it again.

fzdwx: Replaced original response writer with `writerWrapper` in `BeforeInvoke`

mrproliu: You mean the framework not work? Then I think you need to set it as nil if the cast check failure. And if the cast success, you need to call the original hijack method. Right? Otherwise, your framework will still failure for now.

fzdwx: Some websocket server impl:

https://github.com/gorilla/websocket/blob/main/server.go#L175
https://github.com/lxzan/gws/blob/main/upgrader.go#L93

fzdwx: 也许这里可以使用 `http.ResponseController`

fzdwx: https://github.com/golang/go/blob/master/src/net/http/responsecontroller.go#L66

mrproliu: I mean, you cannot break the original plugin. The hijack is only an optional, not requirement.

fzdwx: Sorry, using `http.ResponseController` this object needs to be in 1.20, now i just print log.

mrproliu: A simple way is to let the wrapper implement the Hijack interface, and If the original writer implements the Hijack, you can call the original method. 
Another way is you can create two wrappers, if the case fails, use the current code. otherwise, you create a new write with the Hijack interface.

plugins/mux/serve_interceptor.go
```diff
@@ -63,7 +63,7 @@ func newWriterWrapper(rw interface{}) *writerWrapper {
 	writer := rw.(http.ResponseWriter)
 	hijacker, ok := rw.(http.Hijacker)
 	if !ok {
-		log.Errorf("http.ResponseWriter does not implement http.Hijacker")
+		log.Warnf("http.ResponseWriter does not implement http.Hijacker")
```
mrproliu, line 66: It's unnecessary because the Hijicker is not used for some frameworks. Users will be confused about this log.

plugins/mux/serve_interceptor.go
fzdwx: I tried to add some code, which introduced the `bufio` and `net` packages, and then the following error occurred when building. How to solve it?

```
$WORK/b289/skywalking_enhance_mux_serve_interceptor.go:21:2: could not import bufio (open : no such file or directory)
$WORK/b289/skywalking_enhance_mux_serve_interceptor.go:23:2: could not import net (open : no such file or directory)
```

mrproliu: You cannot import the package which the destination package is imported.

fzdwx: ..., Hijack signature must introduce these two packages.
```go
func (w *writerWrapper) Hijack() (net.Conn, *bufio.ReadWriter, error) {
	rw := w.ResponseWriter
	switch t := rw.(type) {
	case http.Hijacker:
		return t.Hijack()
	default:
		return nil, nil, http.ErrNotSupported
	}
}

```

mrproliu: Then, you can try the second resolution what I mentioned before.

plugins/mux/serve_interceptor.go
```diff
@@ -68,3 +87,16 @@ func (w *writerWrapper) WriteHeader(statusCode int) {
...
+func newWriterWrapperWithHijacker(writer http.ResponseWriter, hijacker http.Hijacker) *writerWrapperWithHijacker {
+	return &writerWrapperWithHijacker{
+		w:        newWriterWrapper(writer),
+		Hijacker: hijacker,
+	}
+}
+
+type writerWrapperWithHijacker struct {
+	w *writerWrapper
+	http.ResponseWriter
```
mrproliu, line 100: The parent wrapper already contains this interface, please delete the duplicate one.

fzdwx: Done.

mrproliu:
Then, please add some plugin test about this. Thanks.

plugins/mux/serve_interceptor.go
```diff
@@ -68,3 +90,18 @@ func (w *writerWrapper) WriteHeader(statusCode int) {
 	w.statusCode = statusCode
 	w.ResponseWriter.WriteHeader(statusCode)
 }
+
+func newWriterWrapperWithHijacker(writer http.ResponseWriter, hijacker http.Hijacker) *writerWrapperWithHijacker {
+	wrapper := newWriterWrapper(writer)
+	return &writerWrapperWithHijacker{
+		ResponseWriter: wrapper,
+		writer:         wrapper,
```
mrproliu, line 98: Why there need to field to reference a same reference?

test/plugins/scenarios/mux/config/excepted.yml
```diff
@@ -16,8 +16,42 @@
...
+            spanType: Entry
+            peer: ''
+            skipAnalysis: false
+            tags:
+              - { key: http.method, value: GET }
+              - { key: url, value: 'localhost:8080/ws' }
+              - { key: status_code, value: '200' }
+      - segmentId: not null
+        spans:
+          - operationName: GET:/health
```
mrproliu, line 40: Please delete the health checker.

test/plugins/scenarios/mux/main.go
```diff
@@ -54,12 +56,50 @@ func health(w http.ResponseWriter, r *http.Request) {
...
+		time.Sleep(3 * time.Second)
 
+		u := url.URL{Scheme: "ws", Host: "localhost:8080", Path: "/ws"}
+		log.Printf("connecting to %s", u.String())
+		c, _, err := websocket.DefaultDialer.Dial(u.String(), nil)
+		if err != nil {
+			log.Fatal("dial:", err)
+		}
+		defer c.Close()
+		c.WriteMessage(websocket.TextMessage, []byte("hello from mux test"))
```
mrproliu, line 100: Could you move this message to the provider handler? The provider handler is the entrypoint.

plugins/mux/serve_interceptor.go
fzdwx: I have tried modifying it to just use `*writerWrapper`

```go
type writerWrapperWithHijacker struct {
*writerWrapper
http.Hijacker
}
```

But when building the following log appears:

```
go build -toolexec="/path/to/agent" -a  .
# github.com/gorilla/mux
$WORK/b289/skywalking_enhance_mux_serve_interceptor.go:55:82: wrapped.writerWrapper undefined (type *skywalking_operatorTypeMuxwriterWrapperWithHijacker has no field or method writerWrapper)
$WORK/b289/skywalking_enhance_mux_serve_interceptor.go:94:3: unknown field writerWrapper in struct literal of type skywalking_operatorTypeMuxwriterWrapperWithHijacker
2024/07/13 08:18:56 exit status 2
```

test/plugins/scenarios/mux/config/excepted.yml
fzdwx: Done

test/plugins/scenarios/mux/main.go
fzdwx: Done.

test/plugins/scenarios/mux/main.go
```diff
@@ -54,11 +58,48 @@ func health(w http.ResponseWriter, r *http.Request) {
...
+
+func connectWs() {
+	u := url.URL{Scheme: "ws", Host: "localhost:8080", Path: "/ws"}
+	log.Printf("connecting to %s", u.String())
+	c, _, err := websocket.DefaultDialer.Dial(u.String(), nil)
+	if err != nil {
+		log.Fatal("dial:", err)
+	}
+	defer c.Close()
+	c.WriteMessage(websocket.TextMessage, []byte("hello from mux test"))
```
mrproliu, line 93: I think when you connect to the ws, there will generate a new exit span? But I can't see it in the excepted file.

mrproliu:
Also, Please add a note in the [CHANEGS.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md)

test/plugins/scenarios/mux/main.go
fzdwx: It does not generate a new exit span.

fzdwx: Do you mean to generate an exit span with commpent id 5005 (http client)?

fzdwx:
> Also, Please add a note in the [CHANEGS.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md?rgh-link-date=2024-07-13T03%3A39%3A01Z)

Done.

test/plugins/scenarios/mux/main.go
mrproliu: Ok, It's another framework we haven't enhanced, right?

fzdwx: Yes, the bottom layer of gorilla does not directly use http.Client, which is the point we are injecting now (*Transport.RoundTrip)

mrproliu: OK, could you help to create other PR to enhance this framework? I think this test code should belong to that plugin test is more reasonable, we can put it in there temporarily. 
Also, Could you add a comment to explain why do need to add this framework in this test case?

fzdwx: Because this framework will convert `http.ResponseWriter` to `http.Hijacker` when receiving a weboscket connection request, if you think it is redundant, I can directly add a line of forced conversion code to the `provider` method.

mrproliu: Oh, I mean write this in the code as a comment. If some develop read this code maybe confuse about it.

fzdwx: Sorry.

fzdwx: Done.

mrproliu (approved):
Thanks. We can merge this PR after all CI success.

PR #190, opened 2024-08-05
fix: when not route is found, the gin operation name is "http.Method:", example: "GET:".

Hair1ossTeenager:
fix: when not route is found, the gin operation name is "http.Method:", example: "GET:".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@ Release Notes.
 
 ### Bug Fixes
 * Fix panic error when root span finished.
+* Fix when not route is found, the gin operation name is "http.Method:", example: "GET:".
 
 0.4.0
 ------------------
--- a/plugins/gin/intercepter.go
+++ b/plugins/gin/intercepter.go
@@ -36,8 +36,12 @@ func (h *ContextInterceptor) BeforeInvoke(invocation operator.Invocation) error 
 		return nil
 	}
 	context := invocation.CallerInstance().(*gin.Context)
+	fullPath := context.FullPath()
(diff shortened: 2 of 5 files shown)


wu-sheng:
What is URL path? If domain(ip) included?

wu-sheng:
@mrproliu do we use e2e tool here? If so, we need a fix.

mrproliu:
> @mrproliu do we use e2e tool here? If so, we need a fix.

Nope, we haven't used the e2e tool here. Just using mock OAP server for plugin testing.

mrproliu:
@Hair1ossTeenager Could you help to add some example code to the plugin test?

Hair1ossTeenager:
> What is URL path? If domain(ip) included?

View source code:
```go
// FullPath returns a matched route full path. For not found routes
// returns an empty string.
//
//	router.GET("/user/:id", func(c *gin.Context) {
//	    c.FullPath() == "/user/:id" // true
//	})
func (c *Context) FullPath() string {
	return c.fullPath
}
```

Hair1ossTeenager:
> @Hair1ossTeenager Could you help to add some example code to the plugin test? Also, the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md) should be updated.

Done.

mrproliu (dismissed):
Cool!

wu-sheng:
CI Fails, need some fix.

mrproliu:
It looks like it's the docker-compose issue, I will create another PR to fix it.

mrproliu (approved):

PR #194, opened 2024-08-21
Add go 1.23 in docker image and remove 1.16-1.18 support.

mrproliu:


Makefile
```diff
@@ -140,6 +140,7 @@ base.image.go1.19 := golang:1.19
 base.image.go1.20 := golang:1.20
 base.image.go1.21 := golang:1.21
 base.image.go1.22 := golang:1.22
+base.image.go1.23 := golang:1.23
```
wu-sheng, line 143: I am good to add 1.23 added. 

Meanwhile, according to go maintenance policy `Each major Go release is supported until there are two newer major releases.`. I think we at least only need to release N-4?

wu-sheng: > go1.19.13 (released 2023-09-06) includes fixes to the go command, and the crypto/tls and net/http packages. See the [Go 1.19.13 milestone](https://github.com/golang/go/issues?q=milestone%3AGo1.19.13+label%3ACherryPickApproved) on our issue tracker for details.

As N-4(1.19) was not released for nearly a year.

wu-sheng: Mostly, 1.18 should be the lowest version we plan to support.

wu-sheng: I refer to https://go.dev/doc/devel/release

mrproliu: Ok. Then we can remove between the 16-19 versions?

wu-sheng: According to release note, we could at least remove 16, 17, 18, 19(this could consider to be kept).

mrproliu: Sure, I will keep the 19 and higher versions.

mrproliu: I will temporarily make this PR a draft to wait for all testing success because all older tests are based on go `1.18`.

wu-sheng:
Do we have a doc to indicate lowest support version? We should update it.

mrproliu:
> Do we have a doc to indicate lowest support version? We should update it.

Yes, I have updated.

wu-sheng:
About native http, the doc says supporting to 1.20.
Do we support that to 1.23? As it is a native lib.

mrproliu:
> About native http, the doc says supporting to 1.20.
> Do we support that to 1.23? As it is a native lib.

Done, all native frameworks have added tests for 1.23.

wu-sheng (approved):

PR #198, opened 2024-08-29
api: add log,metric,span feat to api

ShyunnY:
Current PR is used to design the `toolkit` API, which mainly designs three aspects: trace, metric, log

+ trace: supports attaching events to spans.

+ log: supports manually calling logs

+ metric: supports manually building commonly used metrics

### DIFF
--- a/toolkit/metric/metrics.go
+++ b/toolkit/metric/metrics.go
@@ -1,3 +1,20 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
 package metric
 
 type CounterRef struct{}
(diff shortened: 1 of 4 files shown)


wu-sheng:
@mrproliu Any context about this? I didn't see any discussion about this feature and new APIs.

toolkit/metric/metrics.go
```diff
@@ -0,0 +1,40 @@
+package metric
+
+type CounterRef struct{}
```
mrproliu, line 3: Should we consider making `CounterRef` an `interface`? Then we just need to implement the methods rather than intercept methods?

toolkit/metric/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Get returns the current value of the counter.
+func (c *CounterRef) Get() float64 {
+	return -1
+}
+
+// Inc increments the counter with value.
+func (c *CounterRef) Inc(val float64) {}
+
+type GaugeRef struct {
```
mrproliu, line 13: Same here

toolkit/metric/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Get returns the current value of the gauge.
+func (g *GaugeRef) Get() float64 {
+	return -1
+}
+
+type Histogram struct {
```
mrproliu, line 21: Same here

mrproliu:
> @mrproliu Any context about this? I didn't see any discussion about this feature and new APIs.

It's an OSPP project to adding metrics and logs API in the toolkit.

toolkit/metric/metrics.go
ShyunnY: There is a question: why don't we keep it consistent with the existing code? (like: /toolkit/trace/span)

+ 1. Add interception method for /toolkit/xxx in inst
+ 2. Implement interception method in inst

mrproliu: OK, if keep consistent with existing tracing toolkit, it's fine to me.

mrproliu:
@ShyunnY Looks like you missing some license headers in your created files. Please fix it.

ShyunnY:
> @ShyunnY Looks like you missing some license headers in your created files. Please fix it.

thank you for your guidance, completed。:)

mrproliu (approved):

PR #200, opened 2024-09-01
feat: toolkit span add event impl

ShyunnY:
What type of PR?

> feat: toolkit span add event api impl

What did you do?

> Completed the implementation of the `AddEvent` API in `toolkit/span`. Support attaching Event to Span in manual mode.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ Release Notes.
 ------------------
 #### Features
 
+* support attaching events to span
+
 #### Plugins
 
 #### Bug Fixes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ Release Notes.
 ------------------
 #### Features
 
-* support attaching events to span
+* support attaching events to span in the toolkit.
 
 #### Plugins
 
--- a/plugins/trace-activation/trace/add_event_intercepter.go
+++ b/plugins/trace-activation/trace/add_event_intercepter.go
@@ -15,38 +15,37 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package trace
+package traceactivation
 
-// EventType Defines the type of Span event
-type EventType string
-
-const (
-	// DebugEventType Indicates the event type is "debug"
-	DebugEventType EventType = "debug"
-
-	// InfoEventType Indicates the event type is "info"
-	InfoEventType EventType = "info"
-
-	// WarnEventType Indicates the event type is "warn"
-	WarnEventType EventType = "warn"
-
-	// ErrorEventType Indicates the event type is "error"
-	ErrorEventType EventType = "error"
+import (
+	"github.com/apache/skywalking-go/plugins/core/operator"
+	"github.com/apache/skywalking-go/plugins/core/tracing"
 )
 
-func (*SpanRef) PrepareAsync() {
+type AddEventInterceptor struct {
 }
 
-func (*SpanRef) AsyncFinish() {
+func (h *AddEventInterceptor) BeforeInvoke(invocation operator.Invocation) error {
+	var (
+		defaultEventType  tracing.EventType = "info"
+		defaultEmptyEvent                   = "unknown"
+	)
+
+	span := tracing.ActiveSpan()
+	if span != nil {
+		et := invocation.Args()[0].(tracing.EventType)
+		if len(et) == 0 {
+			et = defaultEventType
+		}
+		event := invocation.Args()[1].(string)
+		if len(event) == 0 {
+			event = defaultEmptyEvent
+		}
+		span.Log(string(et), event)
+	}
+	return nil
 }
 
-// nolint
-func (*SpanRef) SetTag(key string, value string) {
-}
-
-func (*SpanRef) AddLog(...string) {
-}
-
-// AddEvent Add an event of the specified type to SpanRef.
-func (*SpanRef) AddEvent(et EventType, event string) {
+func (h *AddEventInterceptor) AfterInvoke(_ operator.Invocation, _ ...interface{}) error {
+	return nil
 }
--- a/plugins/trace-activation/trace/add_event_intercepter.go
+++ b/plugins/trace-activation/trace/add_event_intercepter.go
@@ -20,26 +20,19 @@ package traceactivation
 import (
 	"github.com/apache/skywalking-go/plugins/core/operator"
 	"github.com/apache/skywalking-go/plugins/core/tracing"
+	"github.com/apache/skywalking-go/toolkit/trace"
 )
 
 type AddEventInterceptor struct {
 }
 
 func (h *AddEventInterceptor) BeforeInvoke(invocation operator.Invocation) error {
-	var (
-		defaultEventType  tracing.EventType = "info"
-		defaultEmptyEvent                   = "unknown"
-	)
-
 	span := tracing.ActiveSpan()
 	if span != nil {
-		et := invocation.Args()[0].(tracing.EventType)
-		if len(et) == 0 {
-			et = defaultEventType
-		}
+		et := invocation.Args()[0].(trace.EventType)
 		event := invocation.Args()[1].(string)
-		if len(event) == 0 {
-			event = defaultEmptyEvent
+		if event == "" {
+			event = defaultEventMsg
 		}
 		span.Log(string(et), event)
 	}
--- a/plugins/trace-activation/trace/async_add_event_intercepter.go
+++ b/plugins/trace-activation/trace/async_add_event_intercepter.go
@@ -1,0 +1,54 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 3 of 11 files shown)


wu-sheng:
Please make sure CI pass.

plugins/trace-activation/trace/add_event_intercepter.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/apache/skywalking-go/plugins/core/operator"
+	"github.com/apache/skywalking-go/plugins/core/tracing"
+)
+
+type AddEventInterceptor struct {
+}
+
+func (h *AddEventInterceptor) BeforeInvoke(invocation operator.Invocation) error {
+	var (
+		defaultEventType  tracing.EventType = "info"
```
mrproliu, line 30: Can we make this a const otherwise variable in function?

test/plugins/scenarios/trace-activation/main.go
```diff
@@ -42,6 +42,7 @@ func consumerHandler(w http.ResponseWriter, r *http.Request) {
 	testContextCarrier()
 	testComponent()
 	testAsyncInCrossGoroutine()
+	testEvent()
```
mrproliu, line 45: Please add a case for the async mode.

plugins/trace-activation/trace/async_add_event_intercepter.go
```diff
@@ -0,0 +1,54 @@
...
+type AsyncAddEventInterceptor struct {
+}
+
+func (h *AsyncAddEventInterceptor) BeforeInvoke(_ operator.Invocation) error {
+	return nil
+}
+
+func (h *AsyncAddEventInterceptor) AfterInvoke(invocation operator.Invocation, _ ...interface{}) error {
+	var (
+		defaultEventType  tracing.EventType = "info"
```
mrproliu, line 34: Can we make this a const otherwise variable in function?

plugins/trace-activation/trace/add_event_intercepter.go
ShyunnY: Sure, I've thought about it. But I'm not sure if creating a new file to make it shareable is the right approach, I'd like to hear your opinion

mrproliu: You could create a new const file to including these.

ShyunnY: Done

ShyunnY:
Currently need to fix plugin testing and write documentation, temporarily blocking CI

ShyunnY:
@mrproliu 

When we use a separate `const.go` file to define constants, there will be a problem: the variable name in the interceptor does not correspond to the const variable name.

We can view the mixed compiled code
**const.go:**
```go
package trace

// Event related constant definitions
const (
        skywalking_operatorVarTracedefaultEventType     = "info"
        skywalking_operatorVarTracedefaultEmptyEvent    = "unknown"
)
```

**add_event.go:**
```go
package trace

type skywalking_operatorTypeTraceAddEventInterceptor struct {
}

func (h *skywalking_operatorTypeTraceAddEventInterceptor) BeforeInvoke(invocation skywalking_operatorTypeOperatorInvocation) error {
        ......
        if span != nil {
                if et == "" {
                        et = skywalking_operatorTypeTracedefaultEventType
                }
                if event == "" {
                        event = skywalking_operatorTypeTracedefaultEmptyEvent
                }
 
        }
        ......
        return nil
}
```

**async_event.go:**
```go
func (h *skywalking_operatorTypeTraceAsyncAddEventInterceptor) AfterInvoke(invocation skywalking_operatorTypeOperatorInvocation, _ ...interface{}) error {
        ......
        if et == "" {
                et = skywalking_operatorTypeTracedefaultEventType
        }
        event := invocation.Args()[1].(string)
        if event == "" {
                event = skywalking_operatorTypeTracedefaultEmptyEvent
        }
        ......
        return nil
}
```

After analysis, this is because when `go-agent` parses AssignStmt, it parses Rhs into Ident, and Ident **lacks corresponding information** to tell `go-agent `how to modify the mapping relationship. 
This is a problem that cannot be solved at this stage, so we still choose to define temporary variables directly in the function stack.

WDYT?

mrproliu:
Please check these code: https://github.com/apache/skywalking-go/blob/main/tools/go-agent/instrument/plugins/rewrite/rewrite.go#L126-L155
The go agent will check all top-level DST nodes, and add there name mapping. Maybe there missing const token check? I think you need to check in there.

ShyunnY:
> Please check these code: https://github.com/apache/skywalking-go/blob/main/tools/go-agent/instrument/plugins/rewrite/rewrite.go#L126-L155 The go agent will check all top-level DST nodes, and add there name mapping. Maybe there missing const token check? I think you need to check in there.

I have done relevant checks, as I said above. When enhancing `add_event.go`,`async_add_event.go`, AST cannot determine the type of LRs and can only perform the default enhancement.

mrproliu:
Can we add a var mapping check when rewriting LRs data?

ShyunnY:
In LRs Ident, it can be _Const_ or _Var_. But we don’t know these, so we can’t map them.

mrproliu:
We have already mapping all var and const, so we know all about it in the rewrite files. Why we can't rewrite about this?

ShyunnY:
For example, the following code:
```go
if et == "" {
et = defaultEventType
}
```
The corresponding AST result is
```json
"Rhs": [
{
"Loc": {
"End": {
"Filename": "main.go",
"Offset": 407,
"Line": 16,
"Column": 25
},
"Start": {
"Filename": "main.go",
"Offset": 391,
"Line": 16,
"Column": 9
}
},
"Name": "defaultEventType",   # <--- we know
"_type": "Ident"
}
```

When we rewrite, we need to determine its type, but the AST lacks the type and can only know the value of `Ident`

ShyunnY:
Therefore, in the rewrite phase, it can only perform the default rewrite and cannot match it. You can try it.

To solve this problem, I rebuilt the code in the interceptor to avoid introducing too many external dependencies. I think this is a good way

mrproliu:
Read the rewrite logic carefully in go agent, when `Ident` type is involved, the rewrite function will look for an existing name in mapping: https://github.com/apache/skywalking-go/blob/main/tools/go-agent/instrument...plugins...rewrite..context.go#L164-L175

And I have tried, you need to check the const and adding them to the var rewrite strategy, that's enough: https://github.com/apache/skywalking-go/blob/main/tools/go-agent/instrument/plugins/rewrite/rewrite.go#L126-L155

ShyunnY:
I added a small feature to the original rewrite, which prevents it from rewriting when the type has SWNoRewrite suffix. This mainly solves the following problems:

+ Since EventType is defined under the `toolkit/trace` package, we need to assert it in the interceptor, but because of the existence of rewrite, it will cause assertion errors during runtime (the type is not an EventType).

So we prevent it from rewriting, which ensures that when mixed compilation, the EventType type of the interceptor can point to the type defined in the `toolkit/trace` api.

> By the way: we can also use this when implementing other functions later, such as metrics and logs!

plugins/trace-activation/trace/add_event_intercepter.go
```diff
@@ -0,0 +1,51 @@
...
+type AddEventInterceptor struct {
+}
+
+func (h *AddEventInterceptor) BeforeInvoke(invocation operator.Invocation) error {
+	var (
+		defaultEventMsg = "unsetEvent"
+	)
+	span := tracing.ActiveSpan()
+	if span != nil {
+		et := invocation.Args()[0].(tracing.EventTypeSWNoRewrite)
```
mrproliu, line 34: I did not understand why you needed to add this one. It same with any third-party framework, you just need to use `trace.EventType`, Then the agent will take care about this type. You declared type `EventTypeSWNoRewrite` is unnecessary.

test/plugins/scenarios/trace-activation/test_service.go
```diff
@@ -112,6 +118,7 @@ func testAsyncInCrossGoroutine() {
 	go func() {
 		s.SetTag("testAsyncTag", "success")
 		s.AddLog("testAsyncLog", "success")
+		s.AddEvent(trace.WarnEventType, "bar")
```
mrproliu, line 121: If you have added the event in the async mode, why the test case cannot found this event?

tools/go-agent/instrument/plugins/rewrite/rewrite.go
```diff
@@ -131,6 +131,13 @@ func (c *Context) rewriteTopLevelNames(files map[*FileInfo]*dst.File) {
 			case *dst.ImportSpec:
 			case *dst.TypeSpec:
 			case *dst.GenDecl:
+				if n.Tok == token.CONST && cursor.Parent() == f {
```
mrproliu, line 134: Could you just combine the token verify between `CONST` and `VAR`? These codes are generally the same.

plugins/trace-activation/trace/consts.go
```diff
@@ -0,0 +1,5 @@
+package traceactivation
```
mrproliu, line 1: Missing LICENSE header in this file.

CHANGES.md
```diff
@@ -6,6 +6,8 @@ Release Notes.
 ------------------
 #### Features
 
+* support attaching events to span
```
mrproliu, line 9:
```suggestion
* support attaching events to span in the toolkit. 
```

test/plugins/scenarios/trace-activation/test_service.go
ShyunnY: You can see it in the logEvent  of the `operationName: testAsyncInCrossGoroutine` test case  :)

mrproliu (approved):

PR #204, opened 2024-09-18
Bump up actions/upload-artifact and actions/download-artifact to v4

CodePrometheus:
After the recent actions CI (such as [CI-1](https://github.com/apache/skywalking-go/actions/runs/10920895152/job/30312587712)) find that v2 does not work, so upgrade to v4 to solve this problem.

### DIFF
--- a/.github/workflows/plugin-tests.yaml
+++ b/.github/workflows/plugin-tests.yaml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-latest, macos-14]
+        os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     steps:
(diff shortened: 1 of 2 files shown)


wu-sheng (dismissed):

mrproliu (approved):

CodePrometheus:
@mrproliu From the recent CI test, not only artifact@v2 is unavailable [test1](https://github.com/apache/skywalking-go/actions/runs/10920895150), but also colima, which macOS-ci depends on, cannot be started [test2](https://github.com/apache/skywalking-go/actions/runs/10924900608), since timeout the action are automatically canceled. I try to increase the corresponding startup parameters to solve the problem, but still find errors during compilation [test3](https://github.com/CodePrometheus/skywalking-go/actions/runs/10933241116/job/30353268180), so I want to temporarily shutdown macOS-ci, and then I will try to locate the problem or find a more stable solution.
cc @wu-sheng

wu-sheng:
Make sense to me.

mrproliu:
Then please rollback all MacOS-related commits and only disable the MacOS test is enough.

PR #205, opened 2024-09-22
feat: support manual reporting of metrics in toolkit

ShyunnY:
In the current PR, I implemented the API defined in the `toolkit/metric` package to support manual construction of metrics to record metric information.


toolkit/metric/api.go
```diff
@@ -18,16 +18,21 @@
 package metric
 
 // NewCounter creates a new counter metrics.
-func NewCounter(name string, opt ...MeterOpt) *CounterRef {
+func NewCounter(name string, opt ...interface{}) *CounterRef {
```
mrproliu, line 21: Why we need change the opt as interface? If it necessary, please add comment in the method to describe what kind of data we need.

test/plugins/scenarios/metric-activation/config/excepted.yml
```diff
@@ -0,0 +1,61 @@
...
+          - 20
+          - 50
+          - 100
+          - 500
+          - 1000
+      - meterId:
+          name: request_duration_with_label
+          tags:
+            - { name: foo, value: bar }
+        histogramBuckets:
```
mrproliu, line 54: Why the histogram only have bucket, but no values?

plugins/toolkit-activation/metric/counter_get_intercepter.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/apache/skywalking-go/plugins/core/metrics"
+	"github.com/apache/skywalking-go/plugins/core/operator"
+)
+
+type CounterGetInterceptor struct{}
+
+func (h *CounterGetInterceptor) BeforeInvoke(invocation operator.Invocation) error {
+	enhanced, ok := invocation.CallerInstance().(operator.EnhancedInstance)
```
mrproliu, line 28: Is any reason of some interceptor implement in the before, some in the after. Should me make it as consistent?

test/plugins/scenarios/metric-activation/config/excepted.yml
ShyunnY: I saw in the error log of CI/plugin-test that I need to add a bucket of `0.0` in `expected.yaml`, which is easy to solve.

At the same time, I also saw a field: `histogramValues `, do I need to configure this? I don't seem to find this field in the documentation [ref doc](https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/plugin-test/#expecteddatayaml). Can we configure it directly?

plugins/toolkit-activation/metric/counter_get_intercepter.go
ShyunnY: Constructing a new real metrics object in the interceptor, in fact, can be done in the **pre/post** phase, it doesn't matter.

Maybe I forgot to check this part, I will fix it and make it consistent in the `AfterInvoke` phase of the interceptor

toolkit/metric/api.go
ShyunnY: Using the `interface{}` type will allow us to handle it better in the interceptor, at least we don't need to do heavy type conversion tasks, which will bring a lot of unnecessary work.

I will strive to write detailed comments to explain to users why and how we should use it.

test/plugins/scenarios/metric-activation/config/excepted.yml
mrproliu: I think you must configure the values in the excepted file, otherwise we cannot known the test is success or not.

toolkit/metric/api.go
wu-sheng: `interface{}` seems a very general type. For a counter, the parameters should be very certain. And building a counter is not a high-frequency operation, I can't see why change this.

mrproliu: But it's not easy understanding in the user, right? 
Can we add an opt structure, and using the enhance instance as the option configuration. Then, user can know which kind of opts they can fill the arguments.

wu-sheng: > But it's not easy understanding in the user, right?

Not only this, more about. How to prevent the wrong implementation in the method? A certain type is a good thing, we should not expose an `interface{}` but no other condition, unless this is the use case.

test/plugins/scenarios/metric-activation/config/excepted.yml
ShyunnY: Done

plugins/toolkit-activation/metric/counter_get_intercepter.go
ShyunnY: Done

toolkit/metric/api.go
ShyunnY: yeah, I used `MeterOpt` as a type constraint to tell users how to use it.
@wu-sheng @mrproliu

plugins/toolkit-activation/metric/new_histogram_intercepter.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/apache/skywalking-go/plugins/core/metrics"
+	"github.com/apache/skywalking-go/plugins/core/operator"
+	"github.com/apache/skywalking-go/toolkit/metric"
+)
+
+type NewHistogramInterceptor struct{}
+
+func (h *NewHistogramInterceptor) BeforeInvoke(_ operator.Invocation) error {
+	return nil
```
mrproliu, line 29: Please using the defined result when interceptor all metrics constructor. It's better to control all object in the same side, if we change the toolkit API or agent, it could have unknown dependency error.

ShyunnY: I don't quite understand this, are you saying that we shouldn't use anonymous function parameters or do you mean something else?

mrproliu: I mean is it possible we can construct the metrics by the plugin?

ShyunnY: This approach may be possible, but we need to consider the following issues:
1. Parameter parsing is necessary in the interceptor. Even if we build it in the plugin, we still need to parse the parameters
2. Using (1) to build a metric instance, if it is implemented in the plugin, will it introduce unnecessary redundant code?
3. In fact, we don’t do anything complicated in the interceptor, as I said above: parse parameters -> build objects. When the dependent `NewXXX` changes, compatibility must be considered
So I think we may not need to change this

mrproliu: Why it's related to the parameter? I just don't want to use the existing instance created from the toolkit API. 
The basic process is: to build and define the result in the `BeforeInvoke`, then the toolkit implementation will not be invoked. The plugin controls all (toolkit)metrics instances, if we want to extend the metrics it's easy to adapt.

mrproliu (approved):

wu-sheng:
Do we have a complete doc? @ShyunnY Would you like to write a blog to introduce the API usage through a blog?

ShyunnY:
yeah, I would write docs and blogs about the latest new features in `toolkit`. I have added it to my schedule and I will finish it as soon as possible. @wu-sheng

PR #207, opened 2024-10-10
Null Pointer Exception Potential

exgalibas:


mrproliu:
Please add more scenario description here, why we need this kind of check?

exgalibas:
> Please add more scenario description here, why we need this kind of check?

because Entity is interface, go interface contains type and data, when type is not nil, the interface will not be nil, but the data can be nil,  show some code example:
type Entity interface {
	Do()
}

type ImpEntity struct {
	Name string
}

func (e *ImpEntity) Do() {
	fmt.Println(e.Name)
}

func main() {
	var i *ImpEntity
	fmt.Println(i == nil)   // true
	Test(i)
}

func Test(e Entity) {
	if e != nil {   // not nil
		e.Do()   // panic
	}
}
so when  tracer.ServiceEntity is nil,  will panic

mrproliu:
Then, can we use the `reflect.ValueOf().IsNil()` to judgment the reference is nil or not? The `reflect` is the more lower level case, we can use the standard way to do this.

mrproliu:
Also, please fix the conflict.

PR #210, opened 2024-12-16
[bugfix] AMQP consumers are not living up to expectations

renyansongno1:
## Interception Point Modification

Since the Consumer function of AMQP091 returns a read-only chan structure, the previous agent logic reads this chan, causing the normal business logic to miss messages.

Therefore, after my consideration, interception can be performed before the message enters this chan to process the message for tracing.

## Other Modification Points

Due to the definition of the AMQP protocol itself, after the client and broker establish a connection, communication is conducted using channels. Throughout the entire communication process, the message does not contain queue name information. Hence, it is necessary to obtain the relationship between ConsumerTag and queue when the Consumer function is called. In subsequent message bodies, the Queue can be mapped out through ConsumerTag.


renyansongno1:
8706EF94-D67A-48DA-AE5A-635B8B73933A

mrproliu:
Hi, @renyansongno1. Thanks for the PR, should we update the plugin test scenario to verify the issue have been fixed or not? Because in the issue mentioned, sounds like we miss some scenario verify?

renyansongno1:
> Hi, @renyansongno1. Thanks for the PR, should we update the plugin test scenario to verify the issue have been fixed or not? Because in the issue mentioned, sounds like we miss some scenario verify?

ok, I will do it soon.

renyansongno1:
@mrproliu I had modify scenario tests, in this tests, only need send more than 1 msg, if we can find these span, the code is work.

mrproliu (approved):

PR #212, opened 2024-12-22
Support span in error status when http code >= 400, and add trace#Error in toolkit API

alexwn:
1、add logic when http status code >= 400, set span isError is true by span.Error method
2、add new method trace.Error for user set san isError to true by self

### DIFF
--- a/plugins/http/server_intercepter.go
+++ b/plugins/http/server_intercepter.go
@@ -60,7 +60,7 @@ func (h *ServerInterceptor) AfterInvoke(invocation operator.Invocation, result .
 	if wrapped, ok := invocation.Args()[0].(*writerWrapper); ok {
 		span.Tag(tracing.TagStatusCode, fmt.Sprintf("%d", wrapped.statusCode))
 		if wrapped.statusCode >= 400 {
-			span.Error(string(wrapped.body))
+			span.Error("")
 		}
 	}
 	span.End()
@@ -70,19 +70,12 @@ func (h *ServerInterceptor) AfterInvoke(invocation operator.Invocation, result .
 type writerWrapper struct {
 	http.ResponseWriter
 	statusCode int
-	body       []byte
 }
 
 func (w *writerWrapper) WriteHeader(statusCode int) {
 	// cache the status code
 	w.statusCode = statusCode
 	w.ResponseWriter.WriteHeader(statusCode)
-}
-
-func (w *writerWrapper) Write(b []byte) (int, error) {
-	// cache the response body
-	w.body = append(w.body, b...)
-	return w.ResponseWriter.Write(b)
 }
 
 func (w *writerWrapper) Hijack() (rwc net.Conn, buf *bufio.ReadWriter, err error) {
--- a/plugins/http/server_intercepter.go
+++ b/plugins/http/server_intercepter.go
@@ -60,7 +60,7 @@ func (h *ServerInterceptor) AfterInvoke(invocation operator.Invocation, result .
 	if wrapped, ok := invocation.Args()[0].(*writerWrapper); ok {
 		span.Tag(tracing.TagStatusCode, fmt.Sprintf("%d", wrapped.statusCode))
 		if wrapped.statusCode >= 400 {
-			span.Error("")
+			span.Error()
 		}
 	}
 	span.End()
--- a/plugins/http/server_intercepter.go
+++ b/plugins/http/server_intercepter.go
@@ -59,6 +59,9 @@ func (h *ServerInterceptor) AfterInvoke(invocation operator.Invocation, result .
 	span := invocation.GetContext().(tracing.Span)
 	if wrapped, ok := invocation.Args()[0].(*writerWrapper); ok {
 		span.Tag(tracing.TagStatusCode, fmt.Sprintf("%d", wrapped.statusCode))
+		if wrapped.statusCode >= 400 {
+			span.Error(string(wrapped.body))
+		}
 	}
 	span.End()
 	return nil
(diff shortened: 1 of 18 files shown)


plugins/http/server_intercepter.go
```diff
@@ -59,6 +59,9 @@ func (h *ServerInterceptor) AfterInvoke(invocation operator.Invocation, result .
 	span := invocation.GetContext().(tracing.Span)
 	if wrapped, ok := invocation.Args()[0].(*writerWrapper); ok {
 		span.Tag(tracing.TagStatusCode, fmt.Sprintf("%d", wrapped.statusCode))
+		if wrapped.statusCode >= 400 {
+			span.Error(string(wrapped.body))
```
wu-sheng, line 63: You should only tag error codes rather than the whole body. It doesn't make sense.

alexwn: done

plugins/http/server_intercepter.go
```diff
@@ -59,6 +59,9 @@ func (h *ServerInterceptor) AfterInvoke(invocation operator.Invocation, result .
 	span := invocation.GetContext().(tracing.Span)
 	if wrapped, ok := invocation.Args()[0].(*writerWrapper); ok {
 		span.Tag(tracing.TagStatusCode, fmt.Sprintf("%d", wrapped.statusCode))
+		if wrapped.statusCode >= 400 {
+			span.Error("")
```
wu-sheng, line 63: @mrproliu I feel we should support Error method with no log message, rather than an empty string.
Error message should a log, right?

alexwn: image
Is that OK?

wu-sheng: You could ref to Java agent, method name is errorOccured.

mrproliu: The error method's parameter is a variable string list, and you can not pass in any methods. Additionally, as I mentioned in [previous comments](https://github.com/apache/skywalking/issues/12882#issuecomment-2556112341), a check for length 0 needs to be added in the Log method.

alexwn: image
image
OK?

wu-sheng: If you have a separate error method, you don't need to support log length=0, aren't you?

alexwn: sorry, I'm lost. Your mean is like this?
image

wu-sheng: Yes, and the original error(logs) method is kept unchanged.

alexwn: Done

mrproliu: Yes, we can do that, but for unified code management, shouldn't the new method still call the Error method (without passing parameters)? I think both options are fine. @wu-sheng WDYT?

wu-sheng: It depends. Error(logs) is different from labeling a span in error status.

test/plugins/scenarios/http/config/excepted.yml
```diff
@@ -57,6 +57,21 @@ segmentItems:
...
               - {key: url, value: 'localhost:8080/provider'}
               - {key: status_code, value: '200'}
+          - operationName: GET:/errortest
+            parentSpanId: 0
+            spanId: 2
+            spanLayer: Http
+            startTime: gt 0
+            endTime: gt 0
+            componentId: 5005
+            isError: false
```
mrproliu, line 67: If we categorize status codes above 400 as errors on the server side, should we maintain the same logic on the client side?

mrproliu:
Also, please update the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md), thanks!

test/plugins/scenarios/http/config/excepted.yml
alexwn: I'm not sure, I think we should.

alexwn:
> Also, please update the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md), thanks!
Done

test/plugins/scenarios/http/config/excepted.yml
mrproliu: Then, please update the same logical on the client side.

alexwn: > Then, please update the same logical on the client side.

Done

wu-sheng:
@alexwn Are you still working on this?

alexwn:
> @alexwn Are you still working on this?
yes, do you have any question? what else should I do?

wu-sheng:
No, there isn't anything to ask. I am just checking whether the CI fix has any blocks.

alexwn:
> No, there isn't anything to ask. I am just checking whether the CI fix has any blocks.

Yes, I saw it. I just changed the code and submitted it

wu-sheng:
Still failing. You may still need to make CI runnable locally, which is much easier to see the error and fix.

alexwn:
> Still failing. You may still need to make CI runnable locally, which is much easier to see the error and fix.

image
image
I think it should be true, Should I change the gin plugin code?

wu-sheng:
If that case is using http component as well, then it is affected by your codes.

alexwn:
> If that case is using http component as well, then it is affected by your codes.

entry span is not affected by my code, exist span is effected by my code. I just change two span's isError=true, exist span passed, entry span failed. So I should only change the exit span' isError=true and leave the entry span original?

wu-sheng:
> > If that case is using http component as well, then it is affected by your codes.
> 
> entry span is not affected by my code, exist span is effected by my code. I just change two span's isError=true, exist span passed, entry span failed. So I should only change the exit span' isError=true and leave the entry span original?

If you want to make changes to another plugin(gin), then you could make them aligned. It depends on you. Ideally, all HTTP RPC relative plugins should follow your new proposal.

wu-sheng:
@alexwn This change seems to impact other plugin tests as well, please fix them.

alexwn:
> @alexwn This change seems to impact other plugin tests as well, please fix them.
kratosv2 except.yml data, one segment expect status_code=200, but actual no status_code, this segment's component is mux(5017), I saw the code of plugin(mux), it have logical to set the status_code, I have no idea how to fix it.
image
img_v3_02ht_4eb602c7-b33d-41d4-89fd-ce9dd92e932g

mrproliu (approved):

PR #213, opened 2024-12-24
Use request.URL.Host instead when request.Host is empty

StoneMoe:
As we discussed in Slack, this will add support for HTTP requests with empty `request.Host`, such as elasticsearch client, which set request URL after a path-only request creation.

Ref: https://the-asf.slack.com/archives/CHQ1L4T54/p1734924623039789


plugins/http/client_intercepter.go
```diff
@@ -30,12 +30,18 @@ type ClientInterceptor struct {
 
 func (h *ClientInterceptor) BeforeInvoke(invocation operator.Invocation) error {
 	request := invocation.Args()[0].(*http.Request)
-	s, err := tracing.CreateExitSpan(fmt.Sprintf("%s:%s", request.Method, request.URL.Path), request.Host, func(headerKey, headerValue string) error {
+	host := request.Host
+	if host == "" {
+		if request.URL != nil {
```
mrproliu, line 35:
```suggestion
	if host == "" && request.URL != nil {
```
I think it can be simplified to a single line.

mrproliu:
Please update the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md) file.

mrproliu:
Reference to the slack link: https://the-asf.slack.com/archives/CHQ1L4T54/p1734924623039789

wu-sheng:
image

Binary is not allowed. Please remove this unexpected file.

mrproliu:
@StoneMoe, please fix the CI failure.

StoneMoe:
> @StoneMoe, please fix the CI failure.

It says:
```
Error: instrument.go:38:9: string `http` has 2 occurrences, make it a constant (goconst)
	return "http"
	       ^
```

Maybe we should fix it by incresing goconst `min-occurrences` options? @mrproliu

mrproliu:
Can you help to check why there say `2 occurrences`, If so, we should add a new const here.

StoneMoe:
> Can you help to check why there say `2 occurrences`, If so, we should add a new const here.

In test file there is a new "http" string I added.

Considering change goconst option `ignore-tests` to `true` instead I think.

mrproliu (approved):

PR #214, opened 2025-02-25
Add agent self-observability

CodePrometheus:
test1


wu-sheng:
FYI, we plan to release 10.2 in two weeks, waiting for the banyandb 0.8 ready. So, if you want 10.2 including this, let me know.

mrproliu (commented):
Do we have ability to disabled the so11y? If user wants more good performance, they can disable it.

CodePrometheus:
> Do we have ability to disabled the so11y? If user wants more good performance, they can disable it.

Perhaps we can add a global switch to control whether the agent reports metrics, or just control so11y this time? If we want to control so11y, should we completely ban so11y-related collection?

wu-sheng:
Is Java agent providing this `disable` feature?

CodePrometheus:
> Is Java agent providing this `disable` feature?

No, javaagent only uses `Config.Meter.ACTIVE` to control whether metrics are reported to the OAP, but the collection process continues to run.

plugins/core/so11y.go
```diff
@@ -0,0 +1,165 @@
...
+	once.Do(func() {
+		instance = &So11y{}
+	})
+	return instance
+}
+
+func (s *So11y) MeasureTracingContextCreation(t *Tracer, isForceSample, isIgnored bool) {
+	if isForceSample {
+		if isIgnored {
+			if s.propagatedIgnoreContextCounter == nil {
```
mrproliu, line 59: Should we instance all metrics when creating the So11y structure, I think all metrics will created multiple times when running in multiple go-routines.

plugins/core/so11y.go
```diff
@@ -0,0 +1,165 @@
...
+	return t
+}
+
+func (t *Tracer) CollectErrorOfPlugin(pluginName string) {
+	if counter, ok := GetSo11y().errorCounterMap.Load(pluginName); ok {
+		if c, ok := counter.(metrics.Counter); ok {
+			c.Inc(1)
+		}
+	} else {
+		counter, _ := GetSo11y().errorCounterMap.LoadOrStore(pluginName, t.NewCounter(
```
mrproliu, line 141: Should we have a lock for this? Because the NewCounter invocation will always generate new metrics and overwrite the older one. https://github.com/apache/skywalking-go/blob/main/plugins/core/metrics.go#L90-L97

mrproliu:
> Then I think we should be good for now. @mrproliu Is there any code impacting the performance/throughput significantly?

Currently, the main finding is that some of the data may be inaccurate (asynchronous registration), but this can be fixed at the code level.

plugins/core/so11y.go
CodePrometheus: I've added double check here, which might now prevent some concurrency issues.

test/plugins/scenarios/so11y/plugin.yml
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+entry-service: http://${HTTP_HOST}:${HTTP_PORT}/so11y
+health-checker: http://${HTTP_HOST}:${HTTP_PORT}/health
+start-script: ./bin/startup.sh
+framework: github.com/gin-gonic/gin
+export-port: 8080
+support-version:
+  - go: 1.19
```
mrproliu, line 23: Can we added more go version check?

mrproliu (commented):
After update the multi go version check in plugin test, its looks good to me.

test/plugins/scenarios/so11y/plugin.yml
CodePrometheus: Well, the test program is simple, I'll add all supported versions of the agent and it won't introduce any costly overhead.

wu-sheng:
@CodePrometheus When you submit PR for the backend, please make sure e2e added and document 0.6 go-agent required.

CodePrometheus:
> @CodePrometheus When you submit PR for the backend, please make sure e2e added and document 0.6 go-agent required.

OK, I'll submit it in the coming days.

PR #216, opened 2025-04-21
Fix: Correctly instrument newproc1 for Go 1.23+ parameter counts (#13188)

zJiaJun:
Fixes #13188

**Problem:**

The runtime instrumentation logic in `tools/go-agent/instrument/runtime/instrument.go` incorrectly checked for a hardcoded number of parameters (3) for the internal `runtime.newproc1` function. This check failed for Go versions 1.23 and later, where the relevant function signature expects 2 parameters for the instrumentation purpose. Consequently, the agent skipped the injection of the necessary code for automatic context propagation when new goroutines were created using `newproc1`, breaking automatic tracing in Go 1.23+ environments.

**Solution:**

This patch addresses the issue by implementing dynamic parameter count checking based on the Go version specified during compilation via the `-lang` flag:

1.  **Introduced Robust Version Checking:** Added a new method `CheckGoVersionGreaterOrEqual(major, minor int) bool` to the `api.CompileOptions` struct in `tools/go-agent/instrument/api/flags.go`. This method reliably parses the `-lang` flag value (e.g., "go1.22", "go1.23", "go2.0") to accurately determine the Go version and compare it against required major and minor versions.
2.  **Dynamic Parameter Check in Runtime Instrument:** Modified the `FilterAndEdit` method within `runtime.Instrument` (`tools/go-agent/instrument/runtime/instrument.go`). It now utilizes the new `opts.CheckGoVersionGreaterOrEqual(1, 23)` method to determine the `expectedParamCount` for `newproc1` (3 if Go version < 1.23, 2 if Go version >= 1.23). The instrumentation logic for `newproc1` now only proceeds if the detected parameter count matches the expected count for the target Go version.

**Impact:**

With this fix, the skywalking-go agent can now correctly identify and instrument the `runtime.newproc1` function on Go 1.23 and later versions, in addition to older versions. This ensures that the automatic context propagation mechanism functions as expected for goroutines created via the standard `go` keyword across all supported Go versions specified by the `-lang` flag during the build process.

**Files Changed:**

*   `tools/go-agent/instrument/api/flags.go`
*   `tools/go-agent/instrument/runtime/instrument.go`

### DIFF
--- a/.github/workflows/plugin-tests.yaml
+++ b/.github/workflows/plugin-tests.yaml
@@ -105,6 +105,7 @@ jobs:
           - go-elasticsearchv8
           - goframe
           - so11y
+          - cross-goroutine
     steps:
       - uses: actions/checkout@v2
         with:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ Release Notes.
 * Fix cannot find file when exec build in test/plugins.
 * Fix not set span error when http status code >= 400
 * Fix http plugin cannot provide peer name when optional Host is empty.
+* Fix Correctly instrument newproc1 for Go 1.23+ parameter counts
 
 #### Issues and PR
 - All issues are [here](https://github.com/apache/skywalking/milestone/219?closed=1)
--- a/go.work
+++ b/go.work
@@ -68,6 +68,7 @@ use (
 	./test/plugins/scenarios/metric-activation
 	./test/plugins/scenarios/goframe
 	./test/plugins/scenarios/so11y
+	./test/plugins/scenarios/cross-goroutine
 
 	./tools/go-agent
 
--- a/test/plugins/scenarios/cross-goroutine/bin/startup.sh
+++ b/test/plugins/scenarios/cross-goroutine/bin/startup.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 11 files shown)


wu-sheng:
@zJiaJun Please refer to [e2e](https://github.com/apache/skywalking-go/blob/main/.github/workflows/plugin-tests.yaml) to add new go version and use some existing plugin tests.

zJiaJun:
> @zJiaJun Please refer to [e2e](https://github.com/apache/skywalking-go/blob/main/.github/workflows/plugin-tests.yaml) to add new go version and use some existing plugin tests.

Okay, I understand. I will update the `.github/workflows/plugin-tests.yaml` workflow to include Go 1.23/1.24 in the test matrix to ensure the fix is validated on the target versions.

mrproliu:
Hi @zJiaJun. Looks like the go agent don’t have the logical to verify the tracer on cross goroutine scenario. Could you help to add one?

zJiaJun:
> Hi @zJiaJun. Looks like the go agent don’t have the logical to verify the tracer on cross goroutine scenario. Could you help to add one?

Hi @mrproliu, thanks for the feedback! That's a valid point. I agree that adding a specific test case to verify tracer propagation across goroutines is important for this fix.

I will work on adding a new test scenario for this. Let's discuss the specific implementation details if needed.

wu-sheng:
Please add ASF APLv2 header to fix CI.

wu-sheng:
irisv12 case seems to have some issues. Could you try to fix that? @zJiaJun

zJiaJun:
> irisv12 case seems to have some issues. Could you try to fix that? @zJiaJun
Hi @wu-sheng 
Following up on the irisv12 test failures:
I found that the iris v12.1.0 framework version specifically causes issues across different Go versions:
With Go 1.19, it fails due to dependencies requiring Go 1.21+ (missing slices package).
Even with Go 1.21, it fails during the Docker build phase at the go mod tidy step, suggesting deeper dependency resolution problems specific to v12.1.0.
The other tested framework versions (v12.2.0 and v12.2.5) appear more stable with Go 1.19 and Go 1.21 respectively in my local tests.
To resolve the CI failures caused by these specific dependency issues, I've removed v12.1.0 from the test matrix in plugin.yml for now.

wu-sheng:
@mrproliu Could you recheck this test failing? Any suggestion?

mrproliu (approved):

PR #218, opened 2025-04-23
Refactor: Replace emptyReporter with DiscardReporter

zJiaJun:
## Purpose
This PR replaces the custom `emptyReporter` implementation in `plugins/core/tracer.go` with the existing `reporter.NewDiscardReporter()` to reduce code duplication and improve maintainability.

## Changes
- Replace `&emptyReporter{}` with `reporter.NewDiscardReporter()` in the `newTracer()` method
- Remove the redundant `emptyReporter` struct and its method implementations

## Motivation and Context
The `emptyReporter` and `DiscardReporter` have identical functionality - they both implement the Reporter interface but don't perform any actual operations. Using the existing `DiscardReporter` implementation reduces code duplication and follows the DRY principle.

### DIFF
--- a/plugins/core/reporter/discard_reporter.go
+++ b/plugins/core/reporter/discard_reporter.go
@@ -39,7 +39,7 @@ func (r *discardReporter) SendLog(log *logv3.LogData) {
 }
 func (r *discardReporter) ConnectionStatus() ConnectionStatus {
 	// do nothing
-	return 0
+	return ConnectionStatusDisconnect
 }
 func (r *discardReporter) Close() {
 	// do nothing
--- a/plugins/core/reporter/discard_reporter_test.go
+++ b/plugins/core/reporter/discard_reporter_test.go
@@ -31,8 +31,8 @@ func TestDiscardReporter(t *testing.T) {
 	r.SendTracing(nil)
 	r.SendMetrics(nil)
 	r.SendLog(nil)
-	if status := r.ConnectionStatus(); status != 0 {
-		t.Errorf("expect 0, actual is %d", status)
+	if status := r.ConnectionStatus(); status != reporter.ConnectionStatusDisconnect {
+		t.Errorf("expect 2, actual is %d", status)
 	}
 	r.Close()
 }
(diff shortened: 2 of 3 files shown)


wu-sheng:
@mrproliu This seems to be a code level optimization only. We could keep this in 0.7?

zJiaJun:
> @mrproliu This seems to be a code level optimization only. We could keep this in 0.7?

@wu-sheng 
I would like to propose adding a Kafka Reporter implementation to SkyWalking-Go to enhance its data reporting capabilities. This feature would allow users to send tracing data to Apache SkyWalking OAP server through Kafka, which is particularly beneficial in high-throughput environments

I would be happy to contribute to this implementation if the proposal is accepted. Please let me know your thoughts on this addition and any specific requirements or considerations I should be aware of.

wu-sheng:
Kafka reporter is fine, as other agents have that. 
The key is dependency management. We need to avoid impacting user app dependency tree.

mrproliu (approved):

wu-sheng:
> Kafka reporter is fine, as other agents have that. 
> The key is dependency management. We need to avoid impacting user app dependency tree. 

It seems we don't have a way to avoid that.
Please go ahead, and let us pick latest kafka go client for this new feature.

zJiaJun:
> @zJiaJun Hi, please fix the UT.

Fixed

zJiaJun:
> It seems we don't have a way to avoid that.
> Please go ahead, and let us pick latest kafka go client for this new feature.

I'll start working on the Kafka Reporter implementation. It may take some time, and I plan to use segmentio/kafka-go v0.4.47 as the client library.

mrproliu (approved):

wu-sheng:
> > It seems we don't have a way to avoid that.
> > Please go ahead, and let us pick latest kafka go client for this new feature.
> 
> I'll start working on the Kafka Reporter implementation. It may take some time, and I plan to use segmentio/kafka-go v0.4.47 as the client library.

Take your time. We have tagged 0.6 release. We have plenty time for 0.7.

PR #219, opened 2025-04-24
fix: The way filepath.Join is used when handling package names doesn't properly account for windows.

navanis:
The use of filepath.Join when handling package names does not properly account for Windows. This issue allows the application to compile on Windows, but it causes interceptors to be bypassed.


wu-sheng:
After this change, is the agent able to instrument go app compiling on Windows? If so,  we should at least add one plugin test on Windows to verify.

wu-sheng:
https://github.com/apache/skywalking-go/actions/runs/14648116633/job/41125628067?pr=219

Please run lint locally, and fix the issue.

navanis:
> After this change, is the agent able to instrument go app compiling on Windows? If so, we should at least add one plugin test on Windows to verify.

So far, it’s working fine on my Windows. I think we could add a plugin test similar to the macOS one, but does it need to be done in this PR?

Makefile
```diff
@@ -165,4 +165,4 @@ $(base.all:%=docker.%) $(base.all:%=docker.push.%): version-check
 
 .PHONY: docker docker.push
 docker: $(base.all:%=docker.%) ## Build docker images for skywalking-go agent and Push docker images to registry
-docker.push: $(base.all:%=docker.push.%)
\ No newline at end of file
+docker.push: $(base.all:%=docker.push.%)
```
wu-sheng, line 168: I think this is a reformat? Could you revert this file change?

wu-sheng:
@mrproliu Could you check the CI env? It seems to have docker compose version error.

mrproliu (approved):

mrproliu (commented):
Could you help update the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md)?

navanis:
> @navanis Could you submit the required and mock GHA control file for Windows compiling through another PR first? As explained, we need to merge that first.

I think it can be added as a job in plugin-tests.yaml to reuse some of the workflows. I'm testing locally and expect to commit over the next few nights.

wu-sheng:
Approved https://github.com/apache/skywalking-go/pull/222

wu-sheng:
You could add the plugin test now.

wu-sheng:
@navanis Are you going to finish this PR soon?

.github/workflows/windows-plugin-test.yaml
```diff
@@ -16,12 +16,136 @@
 
 name: Windows Plugin Tests
 
+concurrency:
+  group: plugins-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 on:
-  workflow_dispatch:
+  pull_request:
```
wu-sheng, line 24: Was this block the task to run? I can't see this task in CI list?

navanis: third party Action is not allowed, I need to replace this

test/plugins/run.sh
```diff
@@ -207,6 +212,11 @@ for framework_version in $frameworks; do
 
   bash ${case_home}/scenarios.sh > ${case_logs}/scenarios.log
   status=$?
+
+  if [[ $os == "windows" ]]; then
+    exit 0
```
mrproliu, line 217: Will the plugin test always succeed on the Windows platform? And will it only run one case?

navanis: The program is designed to run within the local go environment on Windows. Therefore, the outermost loop needs to execute only once to verify the Windows go version and its corresponding framework version. This step serves solely to initiate the service on Windows and always returns 0. The actual verification occurs when the program runs as a container within WSL. I'll integrate the WSL validation with the Windows startup service, just to minimize confusion.

mrproliu (approved):
Thanks! LGTM

PR #227, opened 2025-07-16
Add mutex to fix some data race

mrproliu:
Following https://github.com/apache/skywalking/issues/13363, this PR aims to fix the data race issue.


CodePrometheus (approved):
Thanks!

PR #229, opened 2025-08-14
feat(add trace profile for go agent)

cqhasy:


plugins/core/prof_labels.go
```diff
@@ -0,0 +1,132 @@
...
+	var labels LabelSet
+
+	pprof.ForLabels(ctx, func(key, value string) bool {
+		labels.list = append(labels.list, label{key: key, value: value})
+		return true
+	})
+	return labels
+}
+
+func GetPprofLabelSet() *LabelSet {
```
mrproliu, line 77: Is this method duplicated?

plugins/core/profile.go
```diff
@@ -0,0 +1,343 @@
...
+			task.MaxSamplingCount = parseInt(arg.Value)
+		case "StartTime":
+			task.StartTime = parseInt64(arg.Value)
+		case "CreateTime":
+			task.CreateTime = parseInt64(arg.Value)
+		case "SerialNumber":
+			task.SerialNumber = arg.Value
+		}
+	}
+	fmt.Println("adding profile task:", task)
```
mrproliu, line 138: Please remove all the `fmt.Println` method, using logger to do this.

plugins/core/tracing.go
```diff
@@ -64,6 +68,16 @@ func (t *Tracer) CreateEntrySpan(operationName string, extractor interface{}, op
 	}
 
 	span, _, err := t.createSpan0(ctx, tracingSpan, opts, withRef(ref), withSpanType(SpanTypeEntry), withOperationName(operationName))
+	if err == nil {
+		id := span.GetSegmentID()
+		t.ProfileManager.ToProfile(operationName, id)
```
mrproliu, line 73: Does this mean all the endpoints will be profiling, without any endpoint name checking?

plugins/core/profile.go
```diff
@@ -0,0 +1,343 @@
...
+				}
+				m.TraceProfileTasks[task.TaskId].Status = reporter.Finished
+			}(task)
+			break
+		}
+	}
+
+}
+
+func (m *ProfileManager) StartProfiling(traceSegmentID string) error {
```
mrproliu, line 241: CPU Profiling should be a global work, not related to a specific trace. Because the endpoint will be executing. 
Such as there is one endpoint that receives two identical requests at the same time, `pprof.StartCPUProfile` will be executed twice.

plugins/core/profile.go
```diff
@@ -0,0 +1,343 @@
...
+func (m *ProfileManager) monitor() error {
+	select {
+	// End on timeout
+	case <-time.After(time.Duration(m.currentTask.duration) * time.Minute):
+
+	// End manually
+	case <-m.labelSets[m.currentTask.traceSegmentId].closeChan:
+	}
+	// Stop profiling
+	pprof.StopCPUProfile()
```
mrproliu, line 272: Same here, if there are two requests received, you stopped the CPU profiling when the first endpoint executed, and the second request is not finished. Then, the second request trace profiling is not completed.

plugins/core/profile.go
```diff
@@ -0,0 +1,341 @@
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if m.currentTask == nil {
+		return false
+	}
+	return m.currentTask.endpointName == endpoint
+}
+
+func (m *ProfileManager) IfProfiling() bool {
+	return m.profileEvents.BaseEventStatus[IfProfiling]
```
mrproliu, line 207: Is this will have a data race?

plugins/core/profile.go
```diff
@@ -0,0 +1,341 @@
...
+			task.CreateTime = parseInt64(arg.Value)
+		case "SerialNumber":
+			task.SerialNumber = arg.Value
+		}
+	}
+	m.Log.Info("adding profile task:", task)
+	if _, exists := m.TraceProfileTasks[task.TaskID]; exists {
+		return
+	}
+	endTime := task.StartTime + int64(task.Duration)*60*1000
```
mrproliu, line 157: Please using `time.Time#Add` to add duration.

plugins/core/profile.go
```diff
@@ -0,0 +1,341 @@
...
+		if t.Status == reporter.Reported || t.EndTime < time.Now().Unix() {
+			delete(m.TraceProfileTasks, k)
+		}
+	}
+}
+
+func (m *ProfileManager) tryStartCPUProfiling() {
+	ok, err := m.profileEvents.ExecuteComplexEvent(CouldProfile)
+	if err != nil {
+		m.Log.Errorf("profile event error:%v", err)
```
mrproliu, line 178: What will happen when the `TrySetCurrentTask` executes failure? Will the start profiling keep running? You can consolidate them all into a single method to simplify the logic of task execution.

plugins/core/profile.go
```diff
@@ -0,0 +1,341 @@
...
+		return
+	}
+	t := m.TraceProfileTasks[m.currentTask.taskID]
```
mrproliu, line 181: Actually, this map will only have one data, so there's no need to use the map, right?

plugins/core/reporter/api.go
```diff
@@ -108,11 +108,21 @@ var (
 	ConnectionStatusShutdown   ConnectionStatus = 3
 )
 
+type TaskStatus int
+
+const (
+	Pending TaskStatus = iota
```
mrproliu, line 114: Does this status only work for the trace profiling task? Please add the prefix of the const. In case it conflicts with other concepts.

plugins/core/profile.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+func (m *ProfileManager) monitor() {
+	done := make(chan struct{})
+	var zeroSince time.Time
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
+	go func() {
+		for range ticker.C {
```
mrproliu, line 258: If the task has started, and after one second, the endpoint does not trigger any request. The agent will stop the profiling without reaching the end time? This logic may not be correct. 
Also, if the end time is not reached, and you have stopped the profiling, at this time, the agent receives the request that matches the endpoint. Will it restart the profiling? 
So I think for the simplify, you could just stop the profiling when the end time reached.

mrproliu (approved):

PR #243, opened 2026-04-10
Fix grpc serverstream compat

dingsongjie:


mrproliu:
Thanks, is this a new feature to support some gRPC type? Please fix the CI and update the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md) and [support-plugins.md](https://github.com/apache/skywalking-go/blob/main/docs/en/agent/support-plugins.md)

dingsongjie:
> Thanks, is this a new feature to support some gRPC type? Please fix the CI and update the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md) and [support-plugins.md](https://github.com/apache/skywalking-go/blob/main/docs/en/agent/support-plugins.md)

 Fix gRPC plugin panic on v1.69+ by supporting transport.ServerStream type

mrproliu:
> > Thanks, is this a new feature to support some gRPC type? Please fix the CI and update the [CHANGES.md](https://github.com/apache/skywalking-go/blob/main/CHANGES.md) and [support-plugins.md](https://github.com/apache/skywalking-go/blob/main/docs/en/agent/support-plugins.md)
> 
> Fix gRPC plugin panic on v1.69+ by supporting transport.ServerStream type

Got it, please fix the comment(also Copilot review).

dingsongjie:
CI fails on gRPC < v1.69 because copyFrameworkFS copies all plugin files (including the V2 interceptor that references transport.ServerStream) regardless of whether the Point was activated. In older gRPC versions, this type doesn't exist, causing a compilation error.
Possible fixes:
- Split into two plugin directories (grpc for pre-v1.69, grpc-v169 for v1.69+)
- Or add filtering in copyFrameworkFS to only copy files for activated interceptors
Any preference on which direction to take?

dingsongjie:
@mrproliu

mrproliu:
@dingsongjie Sorry for the late response. You can separate the two plugins on this. Not only the interceptor not suitable, the interceptor point could also not suite.

PR #244, opened 2026-04-23
chore: fix function name in comment to match actual function

purelualight:
fix function name in comment to match actual function


purelualight:
@wu-sheng @mrproliu Hi, Could you please review this PR at your convenience? Thank you very much.

wu-sheng (approved):

PR #249, opened 2026-06-16
Retire EOL Go (1.19–1.23); align images, go.mod floor, and CI to Go 1.24–1.26

wu-sheng:
### What

Retire end-of-life Go 1.19–1.23 and align everything to the currently-supported releases (Go supports only the latest two; 1.26 is current).

- **`go.mod` floor**: bump all source modules + `go.work` to `go 1.24` (was 1.19/1.20).
- **Base images** (`Makefile`): publish **Go 1.24, 1.25, 1.26** only; drop the EOL 1.19–1.23 variants. `Dockerfile` default builder image → `golang:1.24`.
- **CI**: build matrix on **1.24, 1.25, 1.26**; the plugin-test, **windows**-plugin-test, e2e, and unit/instrument/race/lint runners all on **Go 1.24** (the new floor/baseline).
- **Plugin scenarios**: focus the core Go-version-sweep scenarios (`cross-goroutine`, `http`, `runtime_metrics`, `metric/trace/logging-activation`, `pprof`, `so11y`) on **1.24–1.26**.
- **Docs**: `docs/en/setup/docker.md` (image versions) and `docs/en/agent/support-plugins.md` (`tested go v1.24 to v1.26`) + `CHANGES.md`.

### Why

Downstream consumers (Apache SkyWalking OAP e2e, end users) need a Go 1.24+ agent base image to build services whose dependencies (grpc ≥ 1.79, `golang.org/x/crypto` ≥ 0.45) require Go 1.24. Keeping EOL variants/floor around invites stale toolchains and a floor-vs-tested gap, so images, the source floor, CI runners, and plugin-test coverage are aligned to the supported releases.

### ⚠️ Compatibility note

- **Drops the `…-go1.19`…`…-go1.23` image variants** for new commits — anything pinned to those tags (e.g. SkyWalking OAP e2e Go cases) must move to `…-go1.24`+ when it bumps `SW_AGENT_GO_COMMIT`. Previously published EOL-tagged images remain in the registry.
- **Raises the build floor to Go 1.24** — the agent can no longer be built/used with Go < 1.24.

### Validation

Validated locally on **go1.26.1**: the agent builds with floor=1.24, instruments stdlib + runtime via `-toolexec -a`, and an instrumented HTTP service runs with the agent active (profiling/config/report streams) without panic. Full plugin-scenario coverage on 1.24–1.26 runs in CI.

- [x] Update the `CHANGES` log.


test/plugins/scenarios/cross-goroutine/plugin.yml
```diff
@@ -20,9 +20,6 @@ start-script: ./bin/startup.sh
 framework: go
 export-port: 8080
 support-version:
-  - go: 1.19
```
mrproliu, line 23: These versions are for plugin tests. Do these still need to be deleted for testing?

mrproliu (commented):
Please also update the [support-plugins.md](https://github.com/apache/skywalking-go/blob/main/docs/en/agent/support-plugins.md)

test/plugins/scenarios/cross-goroutine/plugin.yml
wu-sheng: Good catch — resolved by going all the way: the module `go.mod` floor (all source modules + `go.work`) is now bumped to `go 1.24`, so the agent no longer claims to build on EOL Go 1.19–1.23. With the floor at 1.24, the base images, the CI build/plugin/e2e runners, and these plugin-test scenarios are all aligned to the supported releases (1.24–1.26), so dropping 1.19–1.23 here is consistent rather than a coverage gap. The unit/instrument/race/lint job also moved to the 1.24 baseline.

wu-sheng:
Thanks for the review @mrproliu. Addressed both points:
- **support-plugins.md**: updated the `tested go vX` ranges to `go v1.24 to go v1.26`.
- **Plugin-test scenario versions**: rather than leave a floor-vs-tested gap, I bumped the `go.mod` floor (all source modules + `go.work`) to `go 1.24` and moved every CI runner (build matrix, plugin-test, windows-plugin-test, e2e, and unit/lint) to Go 1.24+. EOL Go 1.19–1.23 is now fully retired and everything is aligned to the supported releases 1.24–1.26.

Validated locally on go1.26.1: the agent builds with floor=1.24, instruments stdlib+runtime via `-toolexec -a`, and an instrumented service runs with the agent active without panic.

mrproliu (approved):

PR #255, opened 2026-09-19
fix: histogram NoInit panic and RuntimeContext map race

songzhendong:
## Summary
- Fix `newHistogramFromExistingBuckets` allocating with `make([]*histogramBucket, 0, n)` then writing by index (panic on first NoInit histogram conversion at meter collect). Use `make(..., len(buckets))` and keep existing bucket counters.
- Add `RWMutex` to `RuntimeContext` `Get`/`Set`/`clone`, consistent with `CorrelationContext`. Production GLS is per-goroutine and `TakeSnapShot`/`ContinueContext` already clone, so this is primarily defensive against residual shared-map paths that would otherwise fatal with concurrent map read/write.
- Add unit coverage; name the concurrency test `TestRace*` so `make test-race` runs it under the race detector.

## Test plan
- [x] Internal CI on `songzhendong/skywalking-go` (CI / E2E / Plugin / Windows)
- [x] `go test ./plugins/core -run 'TestNewHistogramFromExistingBucketsLen|TestRaceRuntimeContextConcurrentCloneAndSet'`
- [ ] `make test-race` (or `go test -race -run '^TestRace' ./plugins/core`)
- [ ] Official Apache CI on this PR


mrproliu (commented):
Thanks for the bug fix, could you help to update the changes.md?

songzhendong:
Thanks for the review. I've updated `CHANGES.md` under the **0.8.0** section:

```markdown
#### Bug Fixes

* Fix panic when converting NoInit histograms during meter collect.
* Fix concurrent map access in RuntimeContext Get/Set/clone.
```

Putting it in **0.8.0** follows the current changelog layout: `0.7.0` is the latest released version, and `0.8.0` is the open / unreleased section started in #253 (milestone 0.8.0). New fixes that land before the next release go there rather than under `0.7.0`.

mrproliu (approved):

